PR #76, opened 2013-09-17
Fix array validator

burnpanck:
In my eyes, the behaviour of the Array-trait's validator is inconsistent with the documentation: The documentation claims that it will coerce values as if given to the `np.array` function. However, in fact, it rejects everything that is not either an `np.ndarray` instance itself or a Sequence. `np.array` on the other hand is far more lenient.

This is particularly striking with scalars: Python scalars would get converted to zero-dimensional arrays which fulfil most array-requirements perfectly fine!

### DIFF
--- a/traits/tests/test_cythonized_traits.py
+++ b/traits/tests/test_cythonized_traits.py
@@ -17,6 +17,12 @@ def has_no_compiler():
     except:
         return True
 
+def cython_version():
+    if no_cython:
+        return None
+    from Cython.Compiler.Version import version
+    return tuple(int(v) for v in version.split('.'))
+
 SKIP_TEST = has_no_compiler()
 
 class CythonizedTraitsTestCase(unittest.TestCase, UnittestTools):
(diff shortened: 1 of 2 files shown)


burnpanck:
Bump. Any comments?

burnpanck:
I would be willing to implement the changes needed to allow the `dtype` argument of the `Array`/`CArray` traits to specify a (sub-) class of numpy dtypes according to numpy's type hierarchy. I would furthermore propose a new `casting` argument with the same options as [np.can_cast](http://docs.scipy.org/doc/numpy/reference/generated/numpy.can_cast.html), allowing to specify what type conversions are allowed. One could also have specifiers `subok` (allow subclasses of `ndarray`), `order` (memory order), `contiguous` (contiguous memory), similar to [np.array](http://docs.scipy.org/doc/numpy/reference/generated/numpy.array.html). The difference between `Array` and `CArray` could then be in the default arguments for those specifiers, and possibly `Array` could require the validated value to be already an instance of (a subclass of) `ndarray`. The latter, while possibly more systematic, could possibly break existing code, because `Array` did not behave like this before.

If such a PR would include the change of this current PR (to make `Array`/`CArray` accept scalars unless `ndim`/`shape` forbids it), would it be accepted? Otherwise we'll just live with a non-official version of traits.

burnpanck:
Closing, as I have moved on to use traitlets instead.

jonathanrocher:
Sorry to hear it. What made you decide to jump over?

PR #207, opened 2014-10-08
Add `dispatch` and `priority` kwargs to `on_trait_change` decorator.

johntyree:
This PR gets rid of the `*names` argument to the decorator, which never worked, and replaces the magic prefix on the pattern with a dictionary of explicit data and values.

The default values of the kwargs match those of the normal method.


dpinte:
@johntyree ping

johntyree:
@dpinte @itziakos Hi. Sorry, yes. This is ready for review.

itziakos:
:+1:

Issue #209, opened 2014-10-17
Unhashable HasTraits-subclasses cannot be used as intermediate links in extended trait notifiers


burnpanck:
The reason being that the trait-listener machinery for extended trait notification uses a dictionary to track reference cycles, which fails for unhashable objects:

``` python
from traits.api import Any, HasTraits, Instance

class UnhashableHasTraits(HasTraits):
    a = Any
    __hash__ = None

class Container(HasTraits):
    sub = Instance(UnhashableHasTraits)

obj = Container()
obj.on_trait_change(lambda new:None,'sub.a')
# this will cause an exception in the trait-listeners
obj.sub = UnhashableHasTraits()
```

rahulporuri:
With PR https://github.com/enthought/traits/pull/410 merged, I think this is isn't an issue anymore.

PR #210, opened 2014-10-17
Unhashable classes cannot be used as links in extended trait listeners

burnpanck:
This PR includes a test-case for bug #209.


burnpanck:
Closing, as I have moved on to use traitlets instead.

PR #263, opened 2015-08-28
Added an identifier trait that constraints to python ASCII identifiers.

stefanoborini:
Accepts strings in the format "_identifier" or "value", but not invalid identifier such as
"for", "", "0hello".

### DIFF
--- a/traits/trait_types.py
+++ b/traits/trait_types.py
@@ -26,6 +26,7 @@ from __future__ import absolute_import
 
 import datetime
 import operator
+import keyword
 import re
 import sys
 from os.path import isfile, isdir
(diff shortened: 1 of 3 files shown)


corranwebster:
It would be good if this also accepted Python 3 identifiers when running under Python 3.

There is also the `String` trait which allows you to test against a regex (amongst other things) - if done correctly then this may be worthwhile instead of just using `String` traits.

corranwebster:
Little bit of research: in Python 3 strings have an `.isidentifier()` method, which should make the test really easy :)

stefanoborini:
The tests aren't really helpful with this strategy. All it says is that something failed to pass the test, but it doesn't say which one.

corranwebster:
@mdickinson Right, but neither does the regex.

corranwebster:
@stefanoborini If I had to guess the problem is that `u"unicode"` is not a bad identifier value in Python 3.

Similarly `u"\u03C0"` (π) would be valid in Python 3 but not in Python 2, but `u"\u2211"` (∑) is not valid in either.

rkern:
`u"unicode"` is not a bad identifier value in Python 2 either.

traits/trait_types.py
```diff
@@ -365,6 +366,30 @@ class Str ( BaseStr ):
...
+    default_value = Undefined
+
+    # Regexp for identifier recognition
+    _identifier_regex = re.compile("[_A-Za-z][_a-zA-Z0-9]*$")
+
+    def validate ( self, object, name, value ):
+        """ Validates that a specified value is valid for this trait.
+        """
+
+        if (isinstance( value, str ) and
```
rkern: Use `basestring` for this instead of `str`. `unicode` values that otherwise pass the regex are perfectly good values to use.

stefanoborini: I specifically don't want unicode, that's why it's `IdentifierStr`, not `Identifier`.

rkern: I wouldn't add this to `traits`, then. Almost every time I have needed a Python identifier trait, it has been as part of a UI, which necessarily requires `unicode` objects. If your particular use case strictly requires a `str` object, I recommend keeping this trait definition as part of that application.

stefanoborini: Ok. canceling the PR.

PR #276, opened 2016-01-04
Context manager that provisionally sets the ETSConfig toolkit

corranwebster:
With the change in the default ETSConfig toolkit in TraitsUI a pre-existing bug was uncovered where, if the toolkit isn't expressly set, you can:

* attempt to import a preferred toolkit implementation in one library (eg. wx backend for enable)
* have it incidentally (and sometimes sub-optimally, but that's a separate issue) import another library with a different backend preference (eg. qt backend for pyface)
* things go badly from there

This PR adds a new context manager method to the ETSConfig object that allows a library to provisionally choose a backend that other libraries will respect, but then undo that selection if a problem occurs in the context manager's block.  The intent is to wrap the import methods in the various toolkit selection methods with this context manager.

This is not a perfect fix, but will help somewhat with this particular pain point.

### DIFF
--- a/traits/etsconfig/etsconfig.py
+++ b/traits/etsconfig/etsconfig.py
@@ -206,7 +206,7 @@ class ETSConfig(object):
             If the toolkit attribute is already set, then an AttributeError
             will be raised when entering the context manager.
         """
-        if not self.toolkit:
+        if self.toolkit:
             raise AttributeError("ETSConfig toolkit is already set")
         self.toolkit = toolkit
         try:
--- a/traits/etsconfig/etsconfig.py
+++ b/traits/etsconfig/etsconfig.py
@@ -5,6 +5,7 @@
 import sys
 import os
 from os import path
+from contextlib import contextmanager
 
 
 class ETSConfig(object):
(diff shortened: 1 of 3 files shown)


traits/etsconfig/tests/test_etsconfig.py
```diff
@@ -8,7 +8,10 @@
 import sys
 import tempfile
 import time
-import unittest
+if sys.version_info[:2] == (2, 6):
+    import unittest2 as unittest
+else:
+    import unittest
```
pberkes: Elsewhere in the library we use `from traits.testing.unittest_tools import unittest`, which does the same but without repetitions

corranwebster: The `unittest_tools` module imports stuff from `traits`, and `etsconfig` is explictly not dependent on traits so things that it provides (eg. application home directories) can be used in code which doesn't want to use the full traits libraries.

ETSConfig is in traits basically for historical reasons - it used to be part of a lower-level base library that traits depended on, but which was removed as part of the Great Renaming.

rkern: The library code in `traits.etsconfig` can't depend on anything else in `traits`, but the tests don't matter.

corranwebster: Unless someone feels strongly about this, I'd prefer to keep it as-is.

corranwebster:
Following offline discussion with @mdickinson, I'm going to merge this soon.

traits/etsconfig/etsconfig.py
```diff
@@ -184,6 +209,39 @@ def _set_company(self, company):
...
+            If the toolkit attribute is already set, then an ETSToolkitError
+            will be raised when entering the context manager.
+        """
+        if self.toolkit:
+            msg = "ETSConfig toolkit is already set to '{0}'"
+            raise ETSToolkitError(msg.format(self.toolkit))
+        self.toolkit = toolkit
+        try:
+            yield
+        except:
```
pberkes: except Exception?

traits/etsconfig/tests/test_etsconfig.py
pberkes: I don't feel strongly, just a preference

traits/etsconfig/etsconfig.py
corranwebster: Not in this case, want to catch _anything_.

PR #293, opened 2016-05-13
Fix TraitDocumenter for Python 3

corranwebster:
Replace `.next()` method with the `next()` builtin. This allows the traits documenter to work correctly in Python 3.


rkern:
LGTM.

PR #316, opened 2016-06-20
Additional code to locate resources within py2app/py2exe bundles

robmcmullen:
I use py2app & py2exe to make application bundles, and needed this small extra bit of code to locate resources (images, data files, etc.) that get stuffed into the application bundle zip file from the `data_files` parameter in the call to `setup()` in `setup.py`.


corranwebster:
LGTM - merging

PR #323, opened 2016-08-12
(Cython port) Delegates to Event should raise AttributeError for reading.

jvkersch:
This came up while trying to run the traitsui test suite with the Cython port. If you delegate to an `Event` on another class, accessing that attribute should raise an `AttributeError`, not a `DelegationError`. The relevant code in `ctraits.c` is in https://github.com/enthought/traits/blob/master/traits/ctraits.c#L1821-L1824; assuming that `tp->tp_getattro` is not NULL, you essentially always jump over the code that raises the `DelegationError`, letting any `AttributeError`s that come from reading an Event bubble to the surface.


dpinte:
CI issue:

```
python -m pip install --upgrade --no-use-wheel pip
python : DEPRECATION: --no-use-wheel is deprecated and will be removed in the future.  Please use --no-binary :all: instead.
At line:1 char:1
+ python -m pip install --upgrade --no-use-wheel pip
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (DEPRECATION: --... :all: instead.:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
Requirement already up-to-date: pip in c:\python27-x64\lib\site-packages
Command executed with exception: DEPRECATION: --no-use-wheel is deprecated and will be removed in the future.  Please use --no-binary :all: instead.
```

jvkersch:
I saw that in #322, too. Apparently pip started raising that warning, and it turns out that Appveyor lines that start with `- ps` will error even if just a warning is raised.

Probably the best is just to do as pip suggests, and replace `--no-use-wheel` by that other option...

jvkersch:
@dpinte Can you take another look at this? The Appveyor issue was merged in master, and I merged the changes back into this branch. As a result, there are some changes in the diff that don't relate to this PR.

traits/ctraits.pyx
```diff
@@ -1370,15 +1370,8 @@ cdef object getattr_delegate(cTrait trait, CHasTraits obj, object name):
         delegate_attr_name = trait.delegate_attr_name(trait, obj, name)
         tp = type(delegate)
 
-        try:
-            result = getattr(delegate, delegate_attr_name)
-            return result
-        except NotImplementedError as exc:
-            raise exc
-        except Exception as exc:
-            raise DelegationError("The '%.50s' object has no attribute "
```
dpinte: @jvkersch The change in this PR removes the usage of the `DelegationError` for a specific case. I'm not sure if the change in behaviour is desired for all the other cases. I think you'll need to handle and filter `AttributeError`s raised for delegated events to solve the problem, no?

jvkersch: @dpinte I think the other cases never get triggered, but I might be wrong. Looking at the original C code, unless `tp->tp_getattro` is `NULL`, which I think never happens, we always jump over the code that raises the `DelegationError` (via the goto on L1823). On the Cython side, that means we can basically let any exception from the `getattr` call propagate to the caller. Do you know of any cases that would explicitly trigger that `DelegationError`?

jvkersch: To corroborate this, I ran the traits test suite with coverage enabled, and it looks like the whole block from L1826 to L1838 is never exercised. I don't know if that means the code is dead, but I expect so.

schermafbeelding 2016-08-23 om 05 11 18

dpinte: @jvkersch I think we might just go ahead, add a note in the changelog and check with downstream projects.

jvkersch: @dpinte I think I will run with the above ^, and merge this into the Cython branch, if you have no objections.

dpinte:
@jvkersch merged!

jvkersch:
Thanks!

PR #326, opened 2016-08-30
Remove command-line option for setting toolkit

corranwebster:
At present the ETSConfig option looks to see if there is are items `-toolkit` and a toolkit name adjacent in `sys.argv`; if so it removes them and uses the supplied toolkit.  This is completely independent of any other command-line argument processing done by other code, and the processing is done when the toolkit is first requested (which could occur in any number of places, possibly as a side-effect of imports), which may or may not be before any other argument parsing is done.  This is a potential source of subtle bugs.

Additionally, the use of `-toolkit` vs. `--toolkit` may be surprising to some, as it goes against some conventions for command-line argument use.

This PR removes this code and corresponding tests.

This is obviously a backward-incompatible change, although I have never seen the `-toolkit` option used.

Furthermore, this PR shouldn't be merged without the blessing of either @rkern or @mdickinson.

### DIFF
--- a/traits/etsconfig/etsconfig.py
+++ b/traits/etsconfig/etsconfig.py
@@ -245,8 +245,7 @@ class ETSConfig(object):
     def _get_toolkit(self):
         """
         Property getter for the GUI toolkit.  The value returned is, in order
-        of preference: the value set by the application; the value passed on
-        the command line using the '-toolkit' option; the value specified by
+        of preference: the value set by the application; the value specified by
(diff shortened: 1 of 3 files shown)


rkern:
Agreed.

There are a few remaining references to the option in docstrings. `grin -- '-toolkit'` will help find them. It also is documented in the TraitsUI documentation: https://github.com/enthought/traitsui/blob/master/docs/source/traitsui_user_manual/intro.rst#toolkit-selection

rkern:
LGTM!

PR #349, opened 2017-01-14
Update Python versions supported in docs

corranwebster:
Ensure docs match current state from #345


mdickinson:
Thanks, @corranwebster.

PR #353, opened 2017-01-26
BUG: Fix validation of This traits.

### LINKED ISSUE #352: This trait does not accept None, despite the error message says it should
Traits 4.6.0

```
 $ more x.py 
from traits.api import *

class Foo(HasStrictTraits):
    bar = This()

f=Foo()
b=Foo()
f.bar=b
f.bar=None

(venv) $ python x.py 
  File "x.py", line 9, in <module>
    f.bar=None
  File "/Users/sborini/Work/github/simphony/simphony-metaparser/venv/lib/python3.5/site-packages/traits/trait_handlers.py", line 173, in error
    value )
traits.trait_errors.TraitError: The 'bar' trait of a Foo instance must be an instance of the same type as the receiver or None, but a value of None <class 'NoneType'> was specified.
```

I am not sure if the correct behavior is to accept a None (in that case, it's a bug), or not to accept the None (in that case, the error message is confusing and This is inconsistent with Instance)

rkern:
Fixes #352. There was some stale code that "fixes up" `This` traits to act like `Instance(Foo, allow_none=False)` traits. `ctraits.c` has the codepaths to handle `This` traits as-is, so I just removed the "fixup" code.

### DIFF
--- a/traits/has_traits.py
+++ b/traits/has_traits.py
@@ -428,10 +428,6 @@ class MetaHasTraits ( type ):
         # Finish building the class using the updated class dictionary:
         klass = type.__new__( cls, class_name, bases, class_dict )
 
-        # Fix up all self referential traits to refer to this class:
-        for trait in mhto.self_referential:
-            trait.set_validate( ( 11, klass ) )
-
         # Call all listeners that registered for this specific class:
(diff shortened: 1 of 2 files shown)


mdickinson:
Gotta love a PR that only has deletions outside the tests.

:+1:

PR #360, opened 2017-04-06
DOC: add description and example of Either trait type

stevenjkern:
The commonly useful `Either` trait was difficult to discover in the docs and unexplained, so I fixed some errant `n/a`s in the table of predefined traits to make it more obvious to find and added a short section with a description and an example.

### DIFF
--- a/docs/source/traits_user_manual/defining.rst
+++ b/docs/source/traits_user_manual/defining.rst
@@ -287,7 +287,7 @@ the table.
 | CSet             | CSet( [*trait* = None, *value* = None, *items* = True,   |
 |                  | \*\*\ *metadata*] )                                      |
 +------------------+----------------------------------------------------------+
-| n/a              | Constant( *value*[, \*\*\ *metadata*] )                  |
+| Constant         | Constant( *value*[, \*\*\ *metadata*] )                  |
 +------------------+----------------------------------------------------------+
 | Dict, DictStrAny,| Dict( [*key_trait* = None, *value_trait* = None,         |
 | DictStrBool,     | *value* = None, *items* = True, \*\*\ *metadata*] )      |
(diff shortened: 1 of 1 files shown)


docs/source/traits_user_manual/defining.rst
```diff
@@ -460,6 +460,48 @@ on the class on which it was defined. For example::
     >>> # This is also OK, because mary's manager can be an Employee
     >>> mary.manager = fred
 
+.. index:: Either trait
+
+.. _either:
+
+Either
+::::::
+Another predefined trait that merit special explanation is Either. The
```
rkern: `merits`

mdickinson:
Huh. I'd forgotten that the `Either` default was `None` even when `None` wasn't one of the explicit options.

```
>>> class A(HasTraits):
...     missing_value = Either(Float, Str)
... 
>>> a = A()
>>> a.missing_value = None
  File "/Users/mdickinson/.edm/envs/edm/lib/python2.7/site-packages/traits/trait_handlers.py", line 172, in error
    value )
traits.trait_errors.TraitError: The 'missing_value' trait of an A instance must be a float or a string, but a value of None <type 'NoneType'> was specified.
>>> a.missing_value is None
True
```

docs/source/traits_user_manual/defining.rst
```diff
@@ -460,6 +460,48 @@ on the class on which it was defined. For example::
...
+Either trait is intended for attributes that may take a value of more than
+a single trait type, including None. The default value of Either is None.
```
rkern: It would be worth noting that a different default can be provided with a `default=<whatever>` argument.

stevenjkern:
Thanks for the review, @mdickinson and @rkern. I added explanations the `None` is the default value even if `None` is not one of the types explicitly selected, but can be overridden using the `default` argumnet.

mdickinson:
The new wording LGTM. There's some odd-looking markup in the table, though. Here's what I see in the rendered HTML under a recent Sphinx:

screen shot 2017-04-18 at 09 53 29

Note the triple asterisks for the "metadata" in the lines for "Constant" and "Enum".

mdickinson:
I took the liberty of fixing the markup. One remaining niggle: the doc for `Either` doesn't agree with the reality, since in practice `Either` can be happily instantiated with no arguments (and then gives a trait that accepts only `None`). It may be that this should be considered a bug in the implementation rather than the documentation. :-)

mdickinson:
@stevenjkern: If you're happy with my last commit, I think this can be merged.

rkern:
I think I'm happier to document the intended use rather than the full generality of what is, by accident, permitted.

mdickinson:
> I think I'm happier to document the intended use rather than the full generality of what is, by accident, permitted.

Agreed.

PR #382, opened 2017-10-10
Add __dir__ method to HasTraits to allow for IPython tab completion

### LINKED ISSUE #320: Traits attributes don't show up in Jupyter console tab-completion
In IPython / Jupyter 4 consoles, attributes which are `Traits` do not show up in tab-completion.  Regular class- and instance-attributes do.

```
$  jupyter console
Jupyter Console 4.1.1

In [1]: from traits.api import HasTraits, Float

In [2]: class Foo(HasTraits):
   ...:     bar = Float
   ...:     baz = "qux"
   ...: 

In [3]: f = Foo()

In [4]: f.baa = "quux"

In [5]: f.b<tab>
f.baa         f.base_trait  f.baz         

In [5]: f.bar
Out[6]: 0.0
```

stevenjkern:
Closes #320

This PR adds a `__dir__` method to the `HasTraits` class which returns the list of `trait_names` minus the `trait_added` and `trait_modified` attributes which most users will not be looking for when trying to tab complete.

GIF below:
hastraits_tabcomplete

### DIFF
--- a/traits/has_traits.py
+++ b/traits/has_traits.py
@@ -1647,11 +1647,8 @@ class HasTraits ( CHasTraits ):
             the object. This enables tab-completion in IPython.
         """
         trait_names = self.trait_names()
-        blacklist = ('trait_added', 'trait_modified')
-        filtered_trait_names = [
-            name for name in trait_names if name not in blacklist
-        ]
-        return filtered_trait_names
+        method_names = [method for method in self._each_trait_method(self)]
+        return trait_names + method_names
 
     #---------------------------------------------------------------------------
     #  Copies another object's traits into this one:
--- a/traits/has_traits.py
+++ b/traits/has_traits.py
@@ -1648,7 +1648,8 @@ class HasTraits ( CHasTraits ):
         """
         trait_names = self.trait_names()
         method_names = [method for method in self._each_trait_method(self)]
-        return trait_names + method_names
+        class_attrs = vars(self.__class__).keys()
+        return trait_names + method_names + class_attrs
 
     #---------------------------------------------------------------------------
     #  Copies another object's traits into this one:
--- a/traits/has_traits.py
+++ b/traits/has_traits.py
@@ -171,20 +171,20 @@ if sys.version_info[0] >= 3:
         """
         if method[0:2] == '__':
             method = '_%s%s' % ( class_name, method )
-    
+
         result = class_dict.get( method )
         if ((result is not None) and
             is_function_type(result) and
             (getattr( result, 'on_trait_change', None ) is None)):
             return result
-    
+
         for base in bases:
             result = getattr( base, method, None )
             if ((result is not None) and
                 is_unbound_method_type(result) and \
                 (getattr( result, 'on_trait_change', None ) is None)):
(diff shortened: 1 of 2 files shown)


achabotl:
Yay! I'm not sure it's 100% the expected behavior though. Here's tab completion on master:

screen shot 2017-10-10 at 17 07 52

And still on master, if I assign to a trait first. `aa` shows up in the completion.

screen shot 2017-10-10 at 17 11 36

And here it is with this branch:

screen shot 2017-10-10 at 17 08 09

On the one hand, I like having only the names of the defined traits, which works like regular classes. On the other, many users will have a bad surprise once all the Traits methods are gone. The union of both is probably the most reasonable option even though it's noisy. That would be the union of `self.trait_names()` and `dir(self)`, I think. That would significantly complicate the tests though.

rkern:
The problem is that the default completer goes through the bottleneck of `dir()`. Ideally, you'd want it to work progressively: "The user pressed `Tab` just after the `.`, so here's the most common completions with less noise. Oh, they pressed `Tab` again, they must want to see everything, so here's the complete list. Ah, they added `pri` and pressed `Tab`, so lets use the complete list and see everything that starts with `pri` like `print_traits`."

rkern:
But in the meantime, defining a `__dir__()` that provides the current `dir()` output along with the uninitialized trait names would be a net improvement in usability and correctness.

stevenjkern:
@achabotl and @rkern thanks for looking at this. I updated the method to now return the list of trait names and the list of methods. Now the results of tabbing after the dot look like this:
screen shot 2018-05-29 at 1 07 45 pm

traits/has_traits.py
```diff
@@ -1648,7 +1648,8 @@ def __dir__(self):
         """
         trait_names = self.trait_names()
         method_names = [method for method in self._each_trait_method(self)]
-        return trait_names + method_names
+        class_attrs = vars(self.__class__).keys()
```
rkern: This could probably use a unit test.

mdickinson:
Thank you!

mdickinson:
And the law of unintended consequences strikes again. This PR broke envisage. :-( (See enthought/envisage#119).

I'm calling that a bug in Envisage, but if it turns out to be hard to fix in Envisage, we may need to revert this PR. FWIW, I suspect it should be easy to fix in Envisage, so I hope that won't happen.

stevenjkern:
Bwomp bwomp. Didn't see that coming. Does Envisage have a weekly cron job to test against traits/traitsui masters?

_makes note to self: run test suites of downstream ETS packages any time I touch an upstream ETS package_

mdickinson:
> Bwomp bwomp. Didn't see that coming.

Yeah, me neither. :-(

> Does Envisage have a weekly cron job to test against traits/traitsui masters?

There _is_ a weekly cron job, but I'm not sure whether it's running against the master branches or just against the latest pip-installable releases. Let me check.

mdickinson:
Nope, looks like the weekly cron job is simply running with the latest EDM-installable versions of packages, so not against master branches.

PR #431, opened 2019-01-07
CI for documentation

jjenthought:
Let's see if this works..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,9 +34,7 @@ before_install:
   - if [[ ${TRAVIS_OS_NAME} == 'osx' ]]; then ./install-edm-osx.sh; export PATH="${PATH}:/usr/local/bin"; fi
   - edm install -y wheel click coverage
 install:
-  - edm run -- python etstool.py install --runtime=${RUNTIME} || exit
-  - git clone https://github.com/enthought/enthought-sphinx-theme.git
-  - pushd enthought-sphinx-theme && edm run -- pip install -e . && popd
+  - edm run -- python etstool.py install-docs --runtime=${RUNTIME} || exit
 script:
   - edm run -- python etstool.py test --runtime=${RUNTIME} || exit
   - edm run -- python etstool.py docs --runtime=${RUNTIME} || exit
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_install:
   - if [[ ${TRAVIS_OS_NAME} == 'osx' ]]; then ./install-edm-osx.sh; export PATH="${PATH}:/usr/local/bin"; fi
   - edm install -y wheel click coverage
 install:
-  - edm run -- python etstool.py install-docs --runtime=${RUNTIME} || exit
+  - edm run -- python etstool.py install --runtime=${RUNTIME} || exit
 script:
   - edm run -- python etstool.py test --runtime=${RUNTIME} || exit
   - edm run -- python etstool.py docs --runtime=${RUNTIME} || exit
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -36,7 +36,7 @@ before_install:
 install:
   - edm run -- python etstool.py install --runtime=${RUNTIME} || exit
   - git clone https://github.com/enthought/enthought-sphinx-theme.git
-  - pushd enthought-sphinx-theme && edm run -- python setup.py install && popd
+  - pushd enthought-sphinx-theme && edm run -- pip install -e . && popd
```
mdickinson: It looks as though you're installing the theme into the bootstrap environment, when what you really want is to install it in the target EDM environment. The logic for that install should probably go into the `etstool.py` script, rather than directly into the `.travis.yml` config. (That way, it can be used by Appveyor, too.)

jjenthought:
@mdickinson Hopefully this one looks a little more sensible now!

etstool.py
```diff
@@ -268,6 +305,29 @@ def do_in_tempdir(files=(), capture_files=()):
...
+
+    Parameters
+    ----------
+    path : str
+        Path of the directory to be changed into.
+    """
+    if not os.path.exists(path):
+        raise IOError("The path {!r} does not exist", path)
+    elif not os.path.isdir(path):
+        raise NotADirectoryError(
```
mdickinson: `NotADirectoryError` isn't defined in Python 2; I'd like to keep the `etstool.py` script Python-version agnostic for now.

Actually, I'd suggest dropping these checks entirely; `os.chdir` already raises with a helpful error message in these cases, so there's little benefit in the extra checks.
```
>>> import os
>>> os.chdir("README.rst")
NotADirectoryError: [Errno 20] Not a directory: 'README.rst'
>>> os.chdir("bogus")
FileNotFoundError: [Errno 2] No such file or directory: 'bogus'
```

jjenthought:
Ah I didn’t realise that was new to Python 3! One of the drawbacks of relying on Pycharm suggestions without version checking switched on I suppose.

etstool.py
```diff
@@ -123,6 +123,26 @@ def install(runtime, environment):
...
+@click.option('--runtime', default='3.6')
+@click.option('--environment', default=None)
+def install_docs(runtime, environment):
+    """ Install project and dependencies into a clean EDM environment, then
+    install further dependencies required for building documentation.
+
+    """
+    parameters = get_parameters(runtime, environment)
+    args = ['--runtime={}'.format(runtime)]
+    install(args=args, standalone_mode=False)
```
mdickinson: It's a bit awkward to be calling one click function from another, here, and it's not clear to me that the call is correct: are we passing the `--environment` argument on?

I'd suggest refactoring this: we could move the common functionality into a private helper function that's called by both `install` and `install_docs`, or we could turn the `install_docs` functionality into an option (`with_docs`?) to the `install` command.

etstool.py
```diff
@@ -180,6 +216,7 @@ def test_clean(runtime):
     try:
         install(args=args, standalone_mode=False)
         test(args=args, standalone_mode=False)
+        docs(args=args, standalone_mode=False)
```
mdickinson: If we're running the `docs` build here, should we be using `install_docs` rather than `install`?

etstool.py
mdickinson: Alternatively, we could keep things simple and just have one function: have `install` install the Sphinx theme as part of its work. It's already installing Sphinx and its dependencies.

jjenthought: The set-up now is: `install` installs the documentation theme unless you pass a flag requesting it not do that. Arguably it could also not install sphinx if you pass the flag, but I don't want to mess around with the existing script too much!

etstool.py
jjenthought: This is (almost unintentionally) fixed by the changes to install

mdickinson:
@jjenthought This PR is still marked WIP; should that prefix be removed?

jjenthought:
Thanks @mdickinson , I've edited the title!

PR #467, opened 2019-04-15
Test fix: skip test that requires Traits UI if Traits UI is not available.

mdickinson:
This adds a `skipUnless` to a test that requires Traits UI, to ensure that that test doesn't fail if Traits UI is not installed.


mdickinson:
Thanks.

PR #713, opened 2020-01-08
Ignore traits/_version.py instead of traits/version.py

corranwebster:
At some point .gitignore got out of sync with setup.py.


corranwebster:
Nope - this is no

PR #900, opened 2020-02-14
Update changelog for Traits 6.0.0 final release

mdickinson:
Bring in changes from the 6.0.x release branch.


mdickinson:
@corranwebster Ping? Should be a trivial one; this is the final step before I can really call the 6.0 release complete.

PR #913, opened 2020-03-02
Add new implementation of TraitDictObject, per EEP:2

midhun-pm:
This PR changes the implementation of the TraitDictObject as per EEP-2

TraitDictObject is a subclass of TraitDict which extends the inbuilt python dict to fire notifications on changes(addition/removal/modification) made to the items of the dictionary.


traits/trait_dict_object.py
```diff
@@ -24,46 +23,461 @@ class TraitDictEvent(object):
...
+
+        """
+        if key in self:
+            removed = {key: self[key]}
+            result = super().pop(key)
+            self.notifiy(removed=removed)
+            return result
+        else:
+            return default_value
+
```
kitchoi, line 374: The original default value in `TraitDictObject.pop` is `Undefined`. `Undefined` is used as a way to identify when the default value is not provided. If the default value is not provided and the key is not found, we should still get a KeyError:
```
>>> x = {}
>>> x.pop("c")  # fail!
...
KeyError: "c"
```

See #987

traits/trait_dict_object.py
```diff
@@ -98,232 +512,151 @@ class TraitDictObject(dict):
...
+        -------
+        None
 
-        self[key] = value
-        result = self[key]
+        """
+        trait = getattr(self, 'trait', None)
+        name_items = getattr(self, 'name_items', None)
+        object_ref = getattr(self, 'object', None)
+        if trait is None or name_items is None or object_ref is None:
```
kitchoi, line 601: Same here, could we have a test that fails if these lines were removed please?

traits/trait_dict_object.py
```diff
@@ -98,232 +512,151 @@ class TraitDictObject(dict):
...
-            dict.clear(self)
-
-            if self.name_items is not None:
-                self._send_trait_items_event(
-                    self.name_items, TraitDictEvent(removed=removed)
-                )
+        """
+        trait = getattr(self, 'trait', None)
+        object_ref = getattr(self, 'object', None)
+        if object_ref is None or trait is None:
```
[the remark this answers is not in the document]
kitchoi: Resolving this one: I think this block is actually covered by `TestTraitListDictSetPersistence` once the `value_validator` is renamed.

traits/trait_dict_object.py
kitchoi: Checked that after renaming `key_validator` and `value_validator` (so that the code is actually run), removing this block of code still does not cause any tests to fail. I tested this on Python 3.8 and 3.6, investigating.

traits/trait_dict_object.py
```diff
@@ -98,232 +512,151 @@ class TraitDictObject(dict):
...
+            return
 
-        if self.name_items is not None:
-            self._send_trait_items_event(
-                self.name_items, TraitDictEvent(added={key: result})
-            )
+        object = object_ref()
 
-        return result
+        if object is None or not hasattr(self, "trait"):
```
kitchoi, line 606: Looks like we don't need this `hasattr(self, "trait")` anymore. It was there because `__setstate__` used not to restore the `trait` attribute, and `__getstate__` removes `trait`. Here `__setstate__` restores `trait` so the attribute is defined.  For this scenario, `object` is already None so the second half of this clause is not needed.

traits/trait_dict_object.py
kitchoi: See #990 which is targeting `TraitSetObject`, but it addresses more or less the same comment.

midhun-pm:
Verified that this fixes #986

traits/trait_dict_object.py
```diff
@@ -24,46 +23,464 @@ class TraitDictEvent(object):
...
+
+        Raises
+        ------
+        TraitError
+            If the value cannot be validated.
+        """
+        return self._validate(self.value_validator, value,
+                              msg="Each value of the")
+
+    def notifiy(self, added={}, changed={}, removed={}):
```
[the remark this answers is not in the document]
midhun-pm: Went with option 1

traits/trait_dict_object.py
```diff
@@ -24,46 +23,461 @@ class TraitDictEvent(object):
...
-        Old keys and values that were just removed.  If nothing was removed
-        this is an empty dict.
+        Old keys and values that were just removed or None.
+
     """
 
     def __init__(self, added=None, changed=None, removed=None):
-        # Construct new empty dicts every time instead of using a default value
-        # in the method argument, just in case someone gets the bright idea of
-        # modifying the dict they get in-place.
```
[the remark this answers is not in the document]
midhun-pm: Made those args required.

traits/trait_dict_object.py
```diff
@@ -24,46 +23,464 @@ class TraitDictEvent(object):
...
+            added = {key: value}
+
+        result = super().setdefault(key, value)
+
+        if added is not None:
+            self.notifiy(added)
+
+        return result
+
+    def pop(self, key, default_value=None):
```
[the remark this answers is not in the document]
midhun-pm: Fixed with #987

traits/trait_dict_object.py
midhun-pm: Merged #990

traits/trait_dict_object.py
```diff
@@ -24,46 +23,464 @@ class TraitDictEvent(object):
...
+    """
+
+    def __new__(cls, *args, **kwargs):
+        instance = super().__new__(cls)
+        instance.key_validator = None
+        instance.value_validator = None
+        instance.notifiers = []
+        return instance
+
+    def __init__(self, value={}, *, key_validator=None,
```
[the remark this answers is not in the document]
midhun-pm: Changed to use a None default.

traits/trait_dict_object.py
```diff
@@ -24,46 +23,464 @@ class TraitDictEvent(object):
...
+        for key, value in adict.items():
+            key = self.validate_key(key)
+            value = self.validate_value(value)
+            validated_items.append((key, value))
+            if key in self:
+                changed.append((key, self[key]))
+            else:
+                added.append((key, value))
+
+        return dict(validated_items), dict(added), dict(changed)
```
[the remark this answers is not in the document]
midhun-pm: Removed this method, refactored code.

traits/trait_dict_object.py
```diff
@@ -24,46 +23,464 @@ class TraitDictEvent(object):
...
+        added : dict
+            The added items as a dict.
+        changed : dict
+            The changed items as a dict.
+
+        """
+
+        validated_items = []
+        added = []
+        changed = []
```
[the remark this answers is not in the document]
midhun-pm: Not used anymore.

traits/tests/test_trait_dict_object.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+def int_validator(value):
+    if type(value) is int:
+        return value
+    else:
+        raise TraitError
+
+
+class TestTraitList(unittest.TestCase):
```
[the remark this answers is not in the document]
midhun-pm: Sorry about that.

traits/tests/test_trait_dict_object.py
```diff
@@ -0,0 +1,285 @@
...
+        self.assertEqual(self.removed, {})
+
+    def test_deepcopy(self):
+        td = TraitDict({"a": 1, "b": 2}, key_validator=str_validator,
+                       value_validator=int_validator,
+                       notifiers=[self.notification_handler])
+        td_copy = copy.deepcopy(td)
+
+        for itm, itm_cpy in zip(td.items(), td_copy.items()):
+            self.assertTupleEqual(itm_cpy, itm)
```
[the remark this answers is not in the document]
midhun-pm: Hmm, I didn't notice this failure before, changing it .

traits/tests/test_trait_dict_object.py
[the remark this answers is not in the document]
kitchoi: Thanks!

traits/tests/test_trait_dict_object.py
```diff
@@ -0,0 +1,283 @@
...
+    def notification_handler(self, trait_dict, added, changed, removed):
+        self.trait_list = trait_dict
+        self.added = added
+        self.changed = changed
+        self.removed = removed
+
+    def test_init(self):
+        td = TraitDict({"a": 1, "b": 2}, key_validator=str_validator,
+                       value_validator=int_validator)
+        self.assertDictEqual(td, {"a": 1, "b": 2})
```
kitchoi, line 52: Not a request change, just a note: This was changed to `assertEqual` in another PR. The two are equivalent here, and I honestly don't mind which one.

traits/trait_dict_object.py
kitchoi: Resolved by #995

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+        validated_dict = {}
+        added = {}
+        changed = {}
+
+        for key, value in adict.items():
+            validated_key = self.key_validator(key)
+            validated_value = self.value_validator(value)
+
+            if key in self:
+                changed[key] = self[key]
```
kitchoi, line 234: My apologies I overlooked this in the second pass following a change addressing a comment in the first review: this branch became inconsistent with the one below. In the event when the key_validator transform the value, the notification is not consistent with the change on the dict.

Could you please verify this test demonstrates the expected behaviour? If so, please add the test and verify it fails before the fix:
```
    def test_update_with_tranformation(self):

        td = TraitDict(
            {"1": 1, "2": 2},
            key_validator=str,
            notifiers=[self.notification_handler],
        )

        # when
        td.update({1: 2})

        # then
        self.assertEqual(td, {"1": 2, "2": 2})
        self.assertEqual(self.added, {})
        self.assertEqual(self.changed, {"1": 1})
        self.assertEqual(self.removed, {})

```

Then changing these two lines to:
```
            if validated_key in self:
                changed[validated_key] = self[validated_key]
```
should fix the test.

midhun-pm: Verified that this test would fail before this fix. Thanks!

traits/tests/test_trait_dict_object.py
midhun-pm: Changed everything to use `assertEqual`

traits/trait_dict_object.py
```diff
@@ -106,224 +391,141 @@ def __init__(self, trait, object, name, value):
...
+    def __getstate__(self):
+        """ Get the state of the object for serialization.
 
-    def rename(self, name):
-        trait = self.object()._trait(name, 0)
-        if trait is not None:
-            self.name = name
-            self.trait = trait.handler
-        else:
-            logger.debug(
```
mdickinson, line 278: With this change, we no longer need the module-level `logger`: let's remove it.

traits/trait_dict_object.py
```diff
@@ -24,31 +24,25 @@ class TraitDictEvent(object):
...
     ----------
-    added : dict or None
-        New keys and values, or optionally None if nothing was added.
-    changed : dict or None
-        Updated keys and their previous values, or optionally None if nothing
-        was changed.
-    removed : dict or None
-        Old keys and values that were just removed, or optionally None if
-        nothing was removed.
+    added : dict
```
mdickinson, line 27:
```suggestion
    added : dict, optional
```

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+        return value
+
+    def pop(self, key, value=Undefined):
+        """ Remove specified key and return the corresponding
+        value. If key is not found, the default value is returned
+        if given, otherwise KeyError is raised.
+
+        Parameters
+        ----------
+        key : A hashable type.
```
mdickinson, line 279: "type" is misleading here; I suggest "object" in place of "type"

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+            Key to the dict item.
+
+        value : any
+            Value to return if key is absent.
+        """
+
+        if value is Undefined or key in self:
+            removed = super().pop(key)
```
mdickinson, line 287: Is it possible to rework so that the superclass `pop` is called unconditionally?

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+            self.notify(
+                added={},
+                changed={},
+                removed={key: removed}
+            )
+            return removed
+        return value
+
+    def popitem(self):
+        """ Remove and return some(key, value) pair as a tuple. Raise KeyError
```
mdickinson, line 297:
```suggestion
        """ Remove and return some (key, value) pair as a tuple. Raise KeyError
```

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+
+        super().__setitem__(validated_key, validated_value)
+        self.notify(added, changed, removed)
+
+    def __delitem__(self, key):
+        """ Delete the item from the dict indicated by the key.
+
+        Parameters
+        ----------
+        key : A hashable type.
```
mdickinson, line 195:
```suggestion
        key : A hashable object.
```

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+        super().update(validated_dict)
+        self.notify(added=added, changed=changed, removed={})
+
+    def setdefault(self, key, value=None):
+        """ Returns the value if key is present in the dict, else creates the
+        key-value pair and returns the value.
+
+        Parameters
+        ----------
+        key : A hashable type.
```
mdickinson, line 249:
```suggestion
        key : A hashable object.
```

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+    # -- pickle and copy support ----------------------------------------------
+
+    def __getstate__(self):
+        """ Get the state of the object for serialization.
+
+        Notifiers are transient and should not be serialized.
+        """
+        result = self.__dict__.copy()
+        # notifiers are transient and should not be serialized
+        result.pop("notifiers", None)
```
mdickinson, line 324: Why not `del result["notifiers"]`? Can this code ever be called with an object that doesn't have a `notifiers` attribute?

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+        value : any
+            The value to set for the corresponding key.
+
+        """
+
+        removed = {}
+        validated_key = self.key_validator(key)
+        validated_value = self.value_validator(value)
+
+        if key in self:
```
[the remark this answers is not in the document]
kitchoi: Agreed. I spotted this in `update` and overlooked this one :face-palmed:.

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+            changed = {}
+            added = {key: value}
+
+        super().__setitem__(key, value)
+
+        self.notify(added=added, changed=changed, removed={})
+
+        return value
+
+    def pop(self, key, value=Undefined):
```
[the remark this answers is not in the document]
kitchoi: This is how it has been a few versions back. Would it be one for a separate PR?

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+        """ Returns the value if key is present in the dict, else creates the
+        key-value pair and returns the value.
+
+        Parameters
+        ----------
+        key : A hashable type.
+            Key to the item.
+        """
+
+        if key in self:
```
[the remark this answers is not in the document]
kitchoi: This one is tricky because of the possible transformation that could happen on the key. e.g. the given key does not exist on the dict, but after the transformation, it exists. See test `test_setdefault_with_casting`.
If we want to call `super().setdefault` directly, then we would need to check the transformed key also does not exist in the dict so that it actually does `__setitem__`.

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+
+        if key_validator is not None:
+            self.key_validator = key_validator
+
+        if value_validator is not None:
+            self.value_validator = value_validator
+
+        if notifiers is None:
+            notifiers = []
+        self.notifiers = list(notifiers)
```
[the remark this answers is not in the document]
midhun-pm: Don't think there's a need for a shallow copy, removing it.

traits/trait_dict_object.py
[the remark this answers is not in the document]
midhun-pm: We do a un-validated `key in self` check in `setdefault` too,  this confuses me.  Should membership checks be done with transformed or raw keys ?

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+
+        if key in self:
+            changed = {key: self[key]}
+            added = {}
+        else:
+            changed = {}
+            added = {validated_key: validated_value}
+
+        super().__setitem__(validated_key, validated_value)
+        self.notify(added, changed, removed)
```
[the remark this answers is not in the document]
midhun-pm: Sticking to passing by keyword.

traits/trait_dict_object.py
midhun-pm: Nope, I don't think so,  changed to `del`,

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+            removed = self.copy()
+            super().clear()
+            self.notify(added={}, changed={}, removed=removed)
+
+    def update(self, adict):
+        """ Update the values in the dict by the new dict.
+
+        Parameters
+        ----------
+        adict : dict
```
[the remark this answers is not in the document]
midhun-pm: Added support for iterable,  may raise more errors though.

traits/trait_dict_object.py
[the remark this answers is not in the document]
midhun-pm: Do you mean an alias for `Undefined` or create a new singleton class ?

traits/trait_dict_object.py
```diff
@@ -62,8 +56,299 @@ def __init__(self, added=None, changed=None, removed=None):
...
+    return item
+
+
+class TraitDict(dict):
+    """ A subclass of dict that validates keys and values and notifies
+    listeners of any change.
+
+    Parameters
+    ----------
+    value : dict, optional
```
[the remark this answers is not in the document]
midhun-pm: Added support for iterables, but it will now raise a `ValueError` (unpack error) if the iterable does not contain pairs.

traits/trait_dict_object.py
```diff
@@ -57,18 +50,314 @@ def __init__(self, added=None, changed=None, removed=None):
...
+        This simply calls all notifiers provided by the class, if any.
+        The notifiers are expected to have the signature::
+
+            notifier(trait_dict, removed, added, changed)
+
+        Any return values are ignored.
+        """
+
+        for notifier in self.notifiers:
+            notifier(self, removed=removed, added=added, changed=changed)
```
mdickinson, line 160: Let's pass by position instead of by keyword here, for consistency with `TraitList` and `TraitSet`.

Note that the way that we make this call matters, and should be documented: externally-supplied notifiers (including those possibly written in C) could potentially need to know whether they're going to be called with positional arguments or keyword arguments.

```suggestion
            notifier(self, removed, added, changed)
```

traits/trait_dict_object.py
```diff
@@ -57,18 +50,314 @@ def __init__(self, added=None, changed=None, removed=None):
...
+        ----------
+        other : dict or iterable
+            The dict from which values will be updated into this dict.
+        """
+
+        validated_dict = {}
+        added = {}
+        changed = {}
+
+        items = other.items() if isinstance(other, dict) else other
```
mdickinson, line 230: As with the `__init__`, let's use the presence of the `keys` attribute when deciding whether to treat as a mapping or an iterable of pairs.

traits/trait_dict_object.py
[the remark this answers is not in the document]
mdickinson: The proposal in #1020 would allow this to be fixed cleanly. However, while I believe that proposal is the right way forward, the discussion on that proposal is postponed, so let's resolve this for now.

traits/trait_dict_object.py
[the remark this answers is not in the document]
mdickinson: Okay, let's keep it in this PR and consider changing it separately.

traits/trait_dict_object.py
```diff
@@ -57,18 +50,314 @@ def __init__(self, added=None, changed=None, removed=None):
...
+        ----------
+        key : A hashable object.
+            Key to the dict item.
+
+        value : any
+            Value to return if key is absent.
+        """
+
+        should_notify = (value is Undefined or key in self)
+        pop_args = [key, value] if value is not Undefined else [key]
```
mdickinson, line 290: I think this would be clearer spelled out as:
```
if value is Undefined:
    removed = super().pop(key)
else:
    removed = super().pop(key, value)
```

traits/trait_dict_object.py
```diff
@@ -57,18 +50,314 @@ def __init__(self, added=None, changed=None, removed=None):
...
+        removed = super().pop(*pop_args)
+
+        if should_notify:
+            self.notify(
+                removed={key: removed},
+                added={},
+                changed={}
+            )
+            return removed
```
mdickinson, line 299: We can get rid of one of the `return` statements here: we just want to return whatever the `super().pop` call returned.

traits/trait_dict_object.py
```diff
@@ -111,224 +400,142 @@ def __init__(self, trait, object, name, value):
...
+        """
 
-    def __setitem__(self, key, value):
-        trait = getattr(self, "trait", None)
-        if trait is None:
-            dict.__setitem__(self, key, value)
-            return
+        trait = getattr(self, 'trait', None)
+        object_ref = getattr(self, 'object', None)
+        if object_ref is None or trait is None:
```
mdickinson, line 428: Could we add a comment explaining why this check is necessary? And similarly for the `_key_validator`? (I'm currently failing to see why it _is_ necessary, so I don't know what that comment would say ...)

traits/trait_dict_object.py
```diff
@@ -111,224 +400,142 @@ def __init__(self, trait, object, name, value):
...
+            return key
 
-        object = self.object()
-        try:
-            validate = trait.key_trait.handler.validate
-            if validate is not None:
-                key = validate(object, self.name, key)
+        object = object_ref()
```
mdickinson, line 431: Do we need a check here for the case that `object` is `None`?

traits/trait_dict_object.py
mdickinson: Why do we need to pass `self` to the `notifier`? I notice that we're doing this in `TraitList`, but not in `TraitSet`; we should be consistent one way or the other (so no matter what happens here, we'll need to change one of `TraitList` and `TraitSet`).

traits/trait_dict_object.py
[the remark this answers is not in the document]
mdickinson: I'm still confused by why this is necessary; we need at least a comment, and I think it's also worth having a unit test in `test_trait_dict_object`.

traits/trait_dict_object.py
mdickinson: Opened #1027 to make sure we don't forget this.

mdickinson: From discussion in #1028, we want to keep the `self` here (and add it back in `TraitSet`).

traits/trait_dict_object.py
```diff
@@ -57,18 +50,314 @@ def __init__(self, added=None, changed=None, removed=None):
...
+            self.value_validator = value_validator
+
+        if notifiers is None:
+            notifiers = []
+        self.notifiers = notifiers
+
+        if value is None:
+            value = {}
+
+        items = value.items() if isinstance(value, dict) else value
```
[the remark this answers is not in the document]
midhun-pm: Won't it cause a problem if for example we pass in a user defined object that's an iterable with a `keys` attribute but does not have `items()` defined?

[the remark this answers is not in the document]
mdickinson: Maybe, but I suspect this is an unlikely case, and it's the exact same problem that Python's `dict` already has (and so far, few people have complained about that). We want to stick close to the behaviour of the superclass where possible.

traits/trait_dict_object.py
midhun-pm: `__setstate__` sets the `trait` attribute to `None` when de-serialized, so `trait.key_trait.handler.validate` in the following lines, would cause an error. But I guess the `object_ref is None` check can be removed, but `object is None` check should be added.

traits/trait_dict_object.py
midhun-pm: Added the check.

traits/trait_dict_object.py
```diff
@@ -9,46 +9,39 @@
...
-        Old keys and values that were just removed.  If nothing was removed
-        this is an empty dict.
+        Updated keys and their previous values.
     """
 
-    def __init__(self, added=None, changed=None, removed=None):
-        # Construct new empty dicts every time instead of using a default value
-        # in the method argument, just in case someone gets the bright idea of
-        # modifying the dict they get in-place.
+    def __init__(self, removed=None, added=None, changed=None):
```
mdickinson, line 40: This is potentially a backwards-incompatible change: if there's any code out there that creates an event with `TraitDictEvent(added, changed, removed)`, it'll have its meaning subtly changed. I can't _find_ any such code, but that doesn't mean that it doesn't exist.

I see two options: (1) restore the signature to its original, introducing a mismatch with the notification signature, or (2) make `added`, `changed` and `removed` keyword-only arguments, which at least ensures that anyone mis-using `TraitDictEvent` gets a hard error instead of a subtly-changed meaning (which could result in hard-to-diagnose errors). (All uses I can find outside this module already pass by keyword here.)

PR #922, opened 2020-03-05
Add new implementation of TraitSetObject, per EEP:2

midhun-pm:
This PR changes the implementation of the TraitSetObject as per EEP-2

TraitSetObject is a subclass of TraitSet which extends the inbuilt python set to fire notifications on changes(addition/removal) made to the items of the set.


traits/trait_set_object.py
```diff
@@ -8,251 +8,448 @@
...
+        Notes
+        -----
+        Parameters in the notification:
+            removed : set
+                A set containing the removed value.
+            added : set
+                An empty set.
+
+        """
+        validated_values = self.validate(value)
```
kitchoi, line 282: The validator should only be given added values.
See #978

traits/trait_set_object.py
```diff
@@ -8,251 +8,444 @@
...
-    object : weak reference to a HasTraits instance
-        A weak reference to a HasTraits instance that the set has been set
-        as an attribute for.
-    name : str
-        The name of the attribute on the object.
-    name_items : str
-        The name of the items event trait that the trait set will fire when
-        mutated.
+    value : set
+        The value for the set
```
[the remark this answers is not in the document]
midhun-pm: Thanks, removed it.

traits/trait_set_object.py
```diff
@@ -263,35 +460,264 @@ def __reduce_ex__(self, protocol=None):
...
+        Returns
+        -------
+        None
+
+        """
+        is_trait_none = getattr(self, 'trait', None) is None
+        is_name_items_none = getattr(self, 'name_items', None) is None
+        is_object_ref_none = getattr(self, 'object', None) is None
+        if is_trait_none or is_name_items_none or is_object_ref_none:
+            return
```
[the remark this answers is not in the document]
midhun-pm: Had to add this to `TraitList` to get some tests to pass in python 3.8. Need to recreate the equivalent tests here and check.

midhun-pm:
Before merging #978 , I'm not entirely sure if validating only added items is the right thing to do. My gripe is really with the `validate` method in traits, it could return an transformed value, nothing prevents the user from writing a validator that make the api behave strange. 

To illustrate, Consider the following: 

```
from traits.trait_set_object import TraitSet
from traits.api import TraitError


def encrypt_validator(current_set, value):
   "" Convert int to char"""
    return {chr(v) for v in value}

ts = TraitSet(validator=encrypt_validator)

ts.add(1)
ts.remove(1) # Raises a KeyError !

```
Must this behavior be avoided?

kitchoi:
I think the `KeyError` in the example makes sense, i.e. no I don't think we need to avoid it.

FTR, this is the behaviour on master:
```
>>> class Foo(HasTraits):
...     a = Set(Str)
... 
>>> foo = Foo()
>>> foo.a = set()
>>> def to_string(object, name, val):         
...     return str(val)
... 
>>> foo.a.trait.item_trait.handler.validate = to_string
>>> foo.a = set([1, 2, 3])
>>> foo.a
TraitSetObject({'3', '2', '1'})
>>> foo.a.remove(2)
  File "/Users/kchoi/Work/ETS/traits/traits/trait_set_object.py", line 227, in remove
    set.remove(self, value)
KeyError: 2
```

mdickinson:
Agreed that this is the right behaviour. (It's not perfect, but it's better than the alternatives.) It's also consistent with the current released behaviour:

Simple example with Traits 6.0.0:
```
>>> from traits.api import *
>>> class A(HasTraits):
...     foo = Set(CStr)
... 
>>> a = A()
>>> a.foo.add(1)
>>> a.foo.remove(1)
  File "/Users/mdickinson/.edm/envs/testing/lib/python3.6/site-packages/traits/trait_set_object.py", line 223, in remove
    set.remove(self, value)
KeyError: 1
>>> a.foo.remove("1")  # succeeds
```

traits/trait_set_object.py
```diff
@@ -263,35 +460,264 @@ def __reduce_ex__(self, protocol=None):
...
+
+        Notifiers are transient and are restored to the empty list.
+        """
+        name = state.setdefault("name", "")
+        object = state.pop("object", None)
+        if object is not None:
+            state['object'] = ref(object)
+            trait = self.object()._trait(name, 0)
+            if trait is not None:
+                state['trait'] = trait.handler
```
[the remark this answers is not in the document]
midhun-pm: Merged #978

traits/trait_set_object.py
midhun-pm: Merged #978

traits/trait_set_object.py
```diff
@@ -8,251 +8,448 @@
...
+        Notes
+        -----
+        Parameters in the notification:
+            removed : set
+                An empty set.
+            added : set
+                A set containing the added value.
+
+        """
+        validated_values = self.validate(value)
```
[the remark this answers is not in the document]
midhun-pm: Merged #978

traits/trait_set_object.py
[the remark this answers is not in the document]
midhun-pm: It looks like the pickle-unpickle test which had failed for `TraitList` in Python 3.8 also includes a test for `Set`. 
Ref: traits.tests.test_trait_list_dict.TestTraitListDictSetPersistence.test_pickle_whole

I've verified that the test does not fail for python 3.8, removing this block.

traits/trait_set_object.py
```diff
@@ -263,35 +441,255 @@ def __reduce_ex__(self, protocol=None):
...
+        self.object = ref(object)
+        self.name = name
+        self.name_items = None
+        if trait.has_items:
+            self.name_items = name + "_items"
+
+        super().__init__(value, validator=self.validator,
+                         notifiers=[self.notifier])
+
+    def validator(self, current_set, value):
```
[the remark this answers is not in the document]
midhun-pm: Thanks for spotting this!!

traits/trait_set_object.py
```diff
@@ -12,247 +12,425 @@
...
-    removed : dict
-        Old values that were removed. If nothing was removed this is an empty
-        set.
+    added : set
+        New values added to the set or None.
+    removed : set
+        Old values that were removed from the set or None.
     """
 
     def __init__(self, removed=None, added=None):
```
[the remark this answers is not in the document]
midhun-pm: Removed the default args.

traits/trait_set_object.py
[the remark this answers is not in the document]
kitchoi: I tried removing this block of code and no tests fail for Python 3.8

If no tests fail when this block of code is removed, it indicates one of these two:
(1) This block of code is redundant and should be removed.
(2) This block of code is needed but we are lacking the relevant test, and we should add a test

Both of these options require actions. I will investigate option 2.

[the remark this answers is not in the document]
kitchoi: On `self.object` and `self.trait`, looks like the minimum we need is just this:
```
        object = self.object()
        if object is None:
            return
```

This is for two use cases:
1) When the set is pickled and unpickled on its own. In that case, both `trait` and `object` are set to None. A set unpickled on its own cannot fire notifications. Checking either `object is None` or `self.trait is None` is enough.
2) When the object is deep-copied, the set's object reference is lost, but the trait is still there.

I can't think of a scenario where `object` is not None but `trait` is None.

On `self.name_items`, that is because the `List`/`Set`/`Dict` trait has a flag for switching off items notifications altogether:
```
Set(items=False)
```
Honestly I don't know why such a feature exists, but it is there and the check on `if self.name_items is not None` is all over the place.
We are missing the test for `name_items` being None.
Also, I don't think we need the lenient `getattr(self, "name_items", None)`, getting the attribute directly with `self.name_items` should be fine, right?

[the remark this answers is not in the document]
kitchoi: > We are missing the test for name_items being None.

I am on it... PR coming...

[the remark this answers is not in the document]
kitchoi: #990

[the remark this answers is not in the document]
kitchoi: Resolved via #990

traits/trait_set_object.py
```diff
@@ -263,35 +441,252 @@ def __reduce_ex__(self, protocol=None):
...
+                Will be Undefined.
+
+        """
         self.difference_update(value)
         return self
+
+
+class TraitSetObject(TraitSet):
+    """ A specialization of TraitList with a default validator and notifier
+    which provide bug-for-bug compatibility with the TraitsListObject from
```
kitchoi, line 548: Ooops, missed this. All mention of `TraitList` should be changed to `TraitSet`, and `TraitsListObject` to `TraitSetObject`. And there are no bug-for-bug implementations here is there?

kitchoi: And there are a lot of mentions of "list" in this docstring, they should all be changed to "set".

traits/trait_set_object.py
```diff
@@ -263,35 +441,252 @@ def __reduce_ex__(self, protocol=None):
...
+        if trait.has_items:
+            self.name_items = name + "_items"
+
+        super().__init__(value, validator=self._validator,
+                         notifiers=[self.notifier])
+
+    def _validator(self, current_set, value):
+        """ Validates the value by calling the inner trait's validate method
+        and also ensures that the size of the list is within the specified
+        bounds.
```
kitchoi, line 591: Please remove the mention of "ensures that the size of the list is within the specified bounds", likewise for the "Raises" section where this is mentioned again.

traits/trait_set_object.py
midhun-pm: :man_facepalming: Sorry, changed it.

traits/trait_set_object.py
midhun-pm: Done.

mdickinson:
I can take a look.

If #989 goes in, would we want to refactor this similarly?

kitchoi:
`TraitSet` is a lot simpler than `TraitList`. While `TraitSet.validate` does take a set of to-be-added values instead of just individual value, it is just a difference of where the loop over items is. I guess a similar refactoring from validating a set to validating individual item would not have much impact.  However, considering the `validator` VS `item_validator` being a public API, *consistency* across `TraitList`/`TraitSet`/`TraitDict` would be valid reason for doing a similar change.

traits/tests/test_trait_set_object.py
```diff
@@ -0,0 +1,515 @@
...
+        ts.update([True, False, True])
+
+        with self.assertRaises(TraitError):
+            ts.update(["invalid"])
+
+    def test_notification(self):
+        ts = TraitSet({1, 2, 3}, validator=int_validator,
+                      notifiers=[self.notification_handler])
+
+        self.assertSetEqual(ts, {1, 2, 3})
```
mdickinson, line 108: I don't think there's any reason to use `assertSetEqual`; just use `assertEqual` throughout. As the [documentation](https://docs.python.org/3/library/unittest.html#unittest.TestCase.addTypeEqualityFunc) says: "it’s usually not necessary to invoke these methods directly".

traits/trait_set_object.py
```diff
@@ -12,247 +12,425 @@
 import copyreg
 from weakref import ref
 
-from .trait_errors import TraitError
+from traits.trait_errors import TraitError
+
+
+def adapt_trait_validator(trait_validator, name="items"):
```
mdickinson, line 18: Can we get rid of this? The abstraction might make sense if this function were being called once and then the return value being stored somewhere for repeated use, but as it is, the `TraitSetObject._validate` method calls this function every time to create a new function, and then immediately calls that function, so this solution is both inefficient and needlessly indirect.

Instead I'd suggest inlining the code directly in the `TraitSetObject._validate` method (possibly delegating to a separate `TraitSetObject` private helper if that looks cleaner).

traits/trait_set_object.py
```diff
@@ -12,247 +12,425 @@
...
-                        self.name_items, TraitSetEvent(None, added)
-                    )
-        except TraitError as excp:
-            excp.set_prefix("Each element of the")
-            raise excp
+        Notifiers are transient and should not be serialized.
+        """
+        result = self.__dict__.copy()
+        # notifiers are transient and should not be serialized
+        result.pop("notifiers", None)
```
mdickinson, line 201: `del result["notifiers"]` would be simpler. (If `notifiers` isn't in the dictionary, that's an indication of a bug in our logic somewhere, so I don't think we need to guard against that case.)

traits/trait_set_object.py
```diff
@@ -12,247 +12,425 @@
...
+                A set containing the added value.
 
-            if self.name_items is not None:
-                self._send_trait_items_event(
-                    self.name_items, TraitSetEvent(removed)
-                )
+        """
+        validated_values = self.validate(set([value]))
+        if len(validated_values) > 1:
+            raise ValueError("Validator returned {} values "
```
mdickinson, line 233: Please could you add a unit test exercising this path? (Though if we replace the whole-set validator with an item validator, this problem goes away.)

traits/trait_set_object.py
```diff
@@ -12,247 +12,425 @@
...
+        Notes
+        -----
+        Parameters in the notification:
+            removed : set
+                A set containing the removed value.
+            added : set
+                An empty set.
+
+        """
+        super().remove(value)
```
mdickinson, line 266: There's a tricky problem to solve here. Ideally we should report the element that was actually removed from the set, rather than the argument to the `remove` call. E.g., in the snippet below, I'd really want to see `{2}` rather than `{2.0}` for the removed items.
```
Python 3.8.2 (default, Feb 27 2020, 19:56:42) 
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
>>> from traits.api import *
>>> class A(HasTraits):
...     foo = Set(Int)
...         print(f"Added: {event.added}; Removed: {event.removed}")
... 
>>> a = A(foo={1, 2, 3})
>>> a.foo.remove(2.0)
Added: set(); Removed: {2.0}   # Expect {2} rather than {2.0}
```

But I don't know of any way to do this that's clean, reliable and efficient. (The core Python philosophy here is that if the two elements are equal, they should be interchangeable, so you shouldn't care.)

traits/trait_set_object.py
```diff
@@ -12,247 +12,425 @@
...
+                A set containing the removed values.
+            added : set
+                An empty set.
+
+        """
         removed = self.intersection(value)
+
         if len(removed) > 0:
-            set.difference_update(self, removed)
+            super().difference_update(removed)
```
mdickinson, line 313: I think it would be safer to try to pass on the exact same call and parameters to the superclass where possible, so call `super().difference_update(value)` unconditionally, instead of calling `super().difference_updated(removed)` conditionally. (This probably applies to other methods too; this is the method where I noticed it first.)

The idea is that we're extending the behaviour of the built-in `set`, so the code should reflect that: we're doing exactly what the superclass would do, plus some extra (validation/transformation and notification).

traits/trait_set_object.py
```diff
@@ -12,247 +12,425 @@
...
+                An empty set.
+
+        """
+        if not self:
+            return
+        removed = set(self)
+        super().clear()
+        self.notify(removed, set())
 
     def __reduce_ex__(self, protocol=None):
```
mdickinson, line 435: Do we need to keep `__reduce_ex__`? It would be preferable to avoid it and just depend on `__getstate__`, `__setstate__`, `__getnewargs__`, etc., if possible.

traits/trait_set_object.py
mdickinson: See the thread starting at https://mail.python.org/pipermail/python-ideas/2013-September/023143.html for some related discussion.

traits/trait_set_object.py
```diff
@@ -12,247 +12,425 @@
...
+    # TraitSet interface
+    # ------------------------------------------------------------------------
+
+    def validate(self, value):
+        """ Validate the set of values.
+
+        This simply calls the validator provided by the class, if any.
+        The validator is expected to have the signature::
+
+            validator(original_set, value_set)
```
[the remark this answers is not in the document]
midhun-pm: It isn't currently being used, not sure if we want to have it for some unforeseen use-case. @kitchoi Can you think of a reason why it's worth having ?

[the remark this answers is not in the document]
kitchoi: Nope. Neither do I.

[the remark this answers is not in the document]
midhun-pm: Removed

traits/trait_set_object.py
```diff
@@ -12,247 +12,425 @@
...
+    added : set
+        New values added to the set.
+    removed : set
+        Old values that were removed from the set.
     """
 
-    def __init__(self, removed=None, added=None):
-        if removed is None:
-            removed = set()
+    def __init__(self, removed, added):
```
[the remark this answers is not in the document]
midhun-pm: Added default arguments.

traits/trait_set_object.py
midhun-pm: I can only think of an inefficient way:
```
>>> s = {1,2,3}
>>> removed = s.difference(s.difference({2.0}))
>>> removed
{2}
```

traits/trait_set_object.py
midhun-pm: Addressed in #998 .

traits/trait_set_object.py
midhun-pm: Addressed in #998 - No longer used.

midhun-pm:
This code has been refactored in #998, will address any more changes in #998

traits/trait_set_object.py
mdickinson: Yes, if we're okay with inefficient then we can use something obvious like `{x for x in my_set if x == 2.0}` to pick out the relevant element. There's an evil trick efficient way that involves wrapping the target element (in this case `2.0`) in a custom object and then observing the calls to `__eq__` made by the Python set machinery on that custom object, but that fails the "clean" test.

This is probably a case where practicality beats purity, and we should just document the deficiency in the docstring.

mdickinson: Here's some code for the non-clean efficient solution:
```python
class _EqObserver:
    def __init__(self, value):
        self.value = value
        self.matches = []

    def __eq__(self, other):
        equal = (self.value == other)
        if equal:
            self.matches.append(other)
        return equal

    def __hash__(self):
        return hash(self.value)


def set_get(s, obj):
    """
    Look "obj" up in set s.

    Return the actual element of s matching obj if it exists.
    Raise ValueError otherwise.

    Examples
    --------
    >>> set_get({1, 2, 3}, 2.0)
    2
    >>> set_get({1, 2, 3}, 4.0)
        ...
    ValueError: object 4.0 not found in set

    """
    observer = _EqObserver(obj)
    if observer in s:
        return observer.matches[0]
    else:
        raise ValueError(f"object {obj} not found in set")
```

traits/trait_set_object.py
midhun-pm: Moved to `TratSetObject` compatibility layer. But I couldn't find a good way to get rid of it.

traits/trait_set_object.py
midhun-pm: Decided to not do this at the moment in an offline discussion.

traits/trait_set_object.py
```diff
@@ -46,213 +44,446 @@ def __init__(self, removed=None, added=None):
...
-        try:
-            validate = trait.item_trait.handler.validate
-            if validate is not None:
-                value = [validate(object, name, val) for val in value]
+    def __init__(self, value=(), *, item_validator=None, notifiers=None):
+        if item_validator is not None:
+            self.item_validator = item_validator
+        super().__init__(self.item_validator(item) for item in value)
+        if notifiers is not None:
+            self.notifiers = list(notifiers)
```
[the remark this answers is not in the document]
midhun-pm: I think the intention is to convert some other iterable to a list.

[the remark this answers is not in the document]
mdickinson: If that's the case, then the docstring needs fixing. But I don't think we should anticipate use-cases here: if it's enough for all current use-cases to use a list, then let's stick with that.

traits/trait_set_object.py
```diff
@@ -46,213 +44,446 @@ def __init__(self, removed=None, added=None):
...
+        Raises
+        ------
+        KeyError
+            If the value is not found in the set.
+        """
+
+        super().remove(value)
+        self.notify(set([value]), set())
+
+    def update(self, value=()):
```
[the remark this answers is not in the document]
midhun-pm: Thanks! Changed `update`, `difference_update` and `intersection_update` to accept multiple args.

traits/trait_set_object.py
```diff
@@ -46,213 +44,446 @@ def __init__(self, removed=None, added=None):
...
-                set.update(self, added)
-
-                if self.name_items is not None:
-                    self._send_trait_items_event(
-                        self.name_items, TraitSetEvent(None, added)
-                    )
-        except TraitError as excp:
-            excp.set_prefix("Each element of the")
-            raise excp
+        self.update(value)
```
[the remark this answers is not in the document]
midhun-pm: `set.__ior__` and others returns a `NotImplemented` error in python3.6, although the operator seems to work , as in, I can do a `python_set | trait_set` 

There's a way to reproduce the `TypeError` by using `self.update(python_set | self)` which wold have the intended updation effect.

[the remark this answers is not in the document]
mdickinson: To be clear, I didn't mean that you should call `set.__ior__` directly: you should call it through `super`, in the same way we're doing for all other methods we're extending.

If I do this:
```
class TraitSet(set):
    ...
    def __ior__(self, value):
        """ Return self |= value. """
        return super().__ior__(value)
```

then I get the behaviours I expect:
```
>>> from traits.trait_set_object import TraitSet
>>> TraitSet({1, 2, 3})
TraitSet({1, 2, 3})
>>> s = TraitSet({1, 2, 3})
>>> t = {4, 5, 6}
>>> s |= t
>>> s
TraitSet({1, 2, 3, 4, 5, 6})
>>> s |= [1, 2, 3]
TypeError: unsupported operand type(s) for |=: 'TraitSet' and 'list'
```

Now all we need to do is add the validation and notification.

And yes, a direct call to `s.__ior__` with a non-set right-hand side returns `NotImplemented` (which isn't an error - it's a sentinel object that Python uses mostly for figuring out what order to do binary operations in). That's fine: the Python machinery translates it to a `TypeError` at the user level:

```
>>> s.__ior__(2.3)
NotImplemented
>>> s |= 2.3
TypeError: unsupported operand type(s) for |=: 'TraitSet' and 'float'
```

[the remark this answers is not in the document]
midhun-pm: Had tried calling `super().__ior__(value)`, that wasn't working, but I now just learnt that the `NotImplemented` is converted to an error while returning. So calling `super().__ior__(value)` and then returning `self` doesn't work.

Thanks!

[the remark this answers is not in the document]
midhun-pm: It needs to be `return super().__ior__(value)`

traits/trait_set_object.py
```diff
@@ -46,213 +44,446 @@ def __init__(self, removed=None, added=None):
         self.added = added
 
 
-class TraitSetObject(set):
-    """ A subclass of set that fires trait events when mutated.
+# Default item validator for TraitSet.
+
+def accept_anything(item):
```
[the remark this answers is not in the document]
midhun-pm: Issue https://github.com/enthought/traits/issues/1014

traits/trait_set_object.py
```diff
@@ -51,213 +50,518 @@ def __repr__(self):
...
+
+        Returns
+        -------
+        self : TraitSet
+            The updated set.
+        """
+
+        removed = set()
+        added = set()
+        if isinstance(value, (set, frozenset)):
```
mdickinson, line 227: Similarly, can we have a comment indicating why this check is necessary?

traits/trait_set_object.py
```diff
@@ -51,213 +50,518 @@ def __repr__(self):
...
+            A value.
+
+        Returns
+        -------
+        self : TraitSet
+            The updated set.
+        """
+        old_set = self.copy()
+
+        if isinstance(value, (set, frozenset)):
```
mdickinson, line 175: I suggest adding a comment immediately above this indicating why the check is necessary. Even knowing the context, my first thought was that we could just remove this line. It's not good that we're baking in assumptions about what Python currently does here (and those assumptions are fine right now, but may not be in the future), but I'm not seeing an alternative.

traits/trait_set_object.py
```diff
@@ -51,213 +50,518 @@ def __repr__(self):
...
+        self : TraitSet
+            The updated set.
+        """
+
+        removed = set()
+        added = set()
+        if isinstance(value, (set, frozenset)):
+            values = set(value)
+            removed = self.intersection(values)
+            raw_result = values.difference(removed)
```
mdickinson, line 230: Nit: `raw_result` is a bit confusing here: I'm not sure what this is the "result" to. How about `raw_added` and `validated_added` instead?

traits/trait_set_object.py
```diff
@@ -51,213 +50,518 @@ def __repr__(self):
...
+            validated_result = {self.item_validator(item) for item in
+                                raw_result}
+            added = validated_result.difference(self)
```
mdickinson, line 233: I'm not sure I'm following the logic here. Why is the `difference` call needed? Shouldn't `validated_result` already be disjoint from `self`?

mdickinson: Okay, I see; this matters when the validator is transforming values, and the transformed values aren't equal to the originals.

Off-topic for this PR, but I think we should probably document somewhere that we simply don't support this case, and any behaviour should be treated as accidental. It's almost impossible to come up with reasonable and consistent semantics for the various operations.

(In contrast, the case where the item validator may transform the value, but the transformed value compares _equal_ to the original is reasonable: for example, the way that the `Int` trait converts NumPy ints to regular Python ints.)

And now I'm worrying about the case of `Float` traits transforming `Int`s, which _is_ a potential value-converting transformation.

One possible solution is to require/document that `item_validators` should be idempotent. We may then be able to get away with always validating inputs.

This may need more thought, but that shouldn't block this PR.

mdickinson: Opened #1020 for discussion.

traits/trait_set_object.py
midhun-pm: Added the comment.

traits/trait_set_object.py
midhun-pm: Added the comment for now.

PR #933, opened 2020-03-11
Add GitHub PR template

kitchoi:
This PR adds a GitHub PR template to help developers.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,4 @@
 **Checklist**
 - [ ] Tests
-- [ ] Update developer documentation
+- [ ] Update API reference (`docs/source/traits_api_reference`)
+- [ ] Update User manual (`docs/source/traits_user_manual`)
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,3 +2,4 @@
 - [ ] Tests
 - [ ] Update API reference (`docs/source/traits_api_reference`)
 - [ ] Update User manual (`docs/source/traits_user_manual`)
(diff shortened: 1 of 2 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,3 @@
+**Checklist**
+- [ ] Tests
+- [ ] Update developer documentation
```
midhun-pm: Maybe split it up into 
- Update docstrings.
- Update user manual.

I know that I'll help me !

Also do we need : 
- Update type annotations (optional)?

kitchoi: > Maybe split it up into
>
> Update docstrings.
> Update user manual.
> I know that I'll help me !

Good point. Thank you!
Maybe I should say "API reference" (`docs/source/traits_api_reference`)  instead of "developer documentation" or "docstring"?
It is expected that docstring will be updated with the code, because they are co-located. The sphinx doc are the ones that tend to get forgotten.

> Also do we need :
> Update type annotations (optional)?

For my curiosity: What does one need to do for this task?

Anyone can open a PR on traits. It would be good to keep the checklist simple so we don't turn away first-time contributors (or any contributors really).  If it is optional, can we skip this for now?

midhun-pm: This is for [EEP-4](https://github.com/enthought/ets/pull/29) for which an initial attempt has been merged into master Ref: [here](https://github.com/enthought/traits/pull/904)

Updating type annotations is only required if a trait type has been added/modified.

kitchoi: Thanks. I added that now.

kitchoi:
Yeah. Thanks!

PR #912, opened 2020-02-28
Add new implementation of TraitListObject, per EEP:2

midhun-pm:
This PR changes the implementation of the `TraitListObject` as per [EEP-2](https://github.com/enthought/ets/blob/master/docs/source/eeps/eep-2.rst)

`TraitListObject` is a subclass of `TraitList` which extends the inbuilt python `list` to fire notifications on changes(addition/removal/reorganization) made to the contained elements.

PR #942, opened 2020-03-18
[POC] Listener rework for EEP-3

kitchoi:
This PR contains POC code for [EEP 3](https://github.com/enthought/ets/blob/master/docs/source/eeps/eep-3.rst), putting together the `TraitObserverNotifier` and the `TraitList` first class object. I have shamelessly copied code from #912 and #931.

This PR is intended to be closed without merging.

The POC itself is also not finalized yet - I am still working on it.  I am opening this PR now for visibility.  Discussion on the wider design / strategy is very welcome.

(I will *try* to remember using `[skip ci]` in the commit message to avoid taking up shared CI resources)

### DIFF
--- a/poc/interfaces.py
+++ b/poc/interfaces.py
@@ -1,0 +1,8 @@
+import abc
+
+
+class INotifiableObject(abc.ABC):
+
+    @abc.abstractmethod
+    def _notifiers(self, force_create):
+        return []
--- a/poc/observe.py
+++ b/poc/observe.py
@@ -25,20 +25,6 @@ def observe(object, callback, path, remove, dispatch):
             observe(next_object, callback, path.next, remove, dispatch)
 
 
-class INotifier:
-
-    def __call__(self, object, trait_name, old, new):
-        pass
-
-    @property
-    def count(self):
-        return 0
-
-    @count.setter
-    def count(self, value):
-        pass
-
-
 class BaseObserverEvent:
     pass
 
@@ -72,6 +58,16 @@ class TraitObserverNotifier(object):
 
     def __call__(self, object, name, old, new):
         pass
+
+    @property
+    def count(self):
+        if not hasattr(self, "_count"):
+            self._count = 0
+        return self._count
+
+    @count.setter
+    def count(self, value):
+        self._count = value
 
 
 WRAPPERS = {
--- a/poc/observe.py
+++ b/poc/observe.py
@@ -235,6 +235,14 @@ def handle_list_item_changed(event, callback, dispatch, path):
             add_notifiers(item, callback, dispatch, path)
 
 
+class DictValueListener(BaseListener):
+    # TODO: Define this class taking into account
+    # what is going to happen in TraitDictObject
+
+    def __init__(self, notify):
+        self.notify = notify
+
+
 class ListenerPath:
 
     def __init__(self, node, next=None):
@@ -369,7 +377,10 @@ path = ListenerPath(
 
 # Listen to all traits with a metadata 'updated'
 path = ListenerPath(
-    node=FilteredTraitListener(filter=lambda trait: "updated" in trait.__dict__),
+    node=FilteredTraitListener(
+        filter=lambda trait: "updated" in trait.__dict__,
+        notify=True,
+    ),
     next=None,
 )
 
--- a/poc/observe.py
+++ b/poc/observe.py
@@ -106,6 +106,8 @@ def add_notifiers(listener, object, callback, dispatch):
 
 
 def remove_notifiers(listener, object, callback):
+    if object is Undefined:
+        return
     for observer_notifiers in listener.iter_lists_of_notifiers(object):
         for other in observer_notifiers[:]:
             if other.equals(callback):
--- a/poc/observe.py
+++ b/poc/observe.py
@@ -4,6 +4,22 @@ from traits.trait_base import Undefined
 
 
 def observe(object, callback, path, remove, dispatch):
+    """
+    Parameters
+    ----------
+    object : HasTrait or CTrait
+        An object that implements `_notifiers` for returning a list for
+        adding or removing notifiers
+    callback : callable(object, name, old, new)
+        A callable conforming to the notifier signature.
+    path : ListenerPath
+        Path for listening to extended traits.
+    remove : boolean
+        Whether to remove the observer.
+    dispatch : str
+        A string indicating the thread on which notifications should be
+        run.
+    """
 
     listener = path.node
     if listener.notify and remove:
--- a/poc/observe.py
+++ b/poc/observe.py
@@ -61,9 +61,7 @@ class TraitObserverNotifier(object):
 
     @property
     def count(self):
-        if not hasattr(self, "_count"):
-            self._count = 0
-        return self._count
+        return getattr(self, "_count", 0)
 
     @count.setter
     def count(self, value):
--- a/poc/observe.py
+++ b/poc/observe.py
@@ -136,7 +136,7 @@ class BaseListener:
         return getattr(self, "_notify", True)
 
     @notify.setter
-    def notify_setter(self, value):
+    def notify(self, value):
         self._notify = value
 
     def iter_this_targets(self, object):
--- a/poc/observe.py
+++ b/poc/observe.py
@@ -19,33 +19,6 @@ class INotifier:
         pass
 
 
-class INotifiableObject:
-
-    def _get_notifiers(self):
-        return []
-
-
-class Trait(INotifiableObject):
-
-    def _get_notifiers(self):
-        return self._notifiers(True)
-
-
-class HasTrait(INotifiableObject):
-
-    def _get_notifiers(self):
-        return self._notifiers(True)
-
-
-class TraitListObject(INotifiableObject):
-
-    def _get_notifiers(self):
-        return []
-
-    def _add_notifier(self):
-        pass
-
-
 class BaseObserverEvent:
     pass
 
--- a/poc/observe.py
+++ b/poc/observe.py
@@ -1,4 +1,4 @@
-
+from functools import partial
 
 
 def observe(has_trait, callback, path, remove, dispatch):
@@ -147,16 +147,27 @@ class FilteredTraitListener(BaseListener):
                 yield object._trait(name, 2)._notifiers(True)
 
 
-class SimpleTraitListener(BaseListener):
+class NamedTraitListener(BaseListener):
 
-    def __init__(self, name, notify):
+    def __init__(self, name, notify, optional):
         self.name = name
         self.notify = notify
+        self.optional = optional
 
     def iter_lists_of_notifiers(self, object):
         trait = object.trait(name=self.name)
         if trait is not None:
             yield object._trait(self.name, 2)._notifiers(True)
+        else:
+            # The trait is not defined.
+            if not self.optional:
+                raise ValueError(
+                    "Trait name {!r} is not defined".format(self.name))
+
+
+OptionalTraitListener = partial(NamedTraitListener, optional=True)
+
+RequiredTraitListener = partial(NamedTraitListener, optional=False)
 
 
 class ListItemListener(BaseListener):
@@ -165,8 +176,12 @@ class ListItemListener(BaseListener):
         self.notify = notify
 
     def iter_lists_of_notifiers(self, object):
-        # Not sure what to do here.
-        pass
+        # object should be a TraitListObject
+        yield object.trait._notifiers(True)
+
+    def iter_next_objects(self, object):
+        # object should be a TraitListObject
+        return iter(object)
 
 
 class ListenerPath:
@@ -179,9 +194,9 @@ class ListenerPath:
 # "inst.attr"
 # fire if inst is changed, or inst.attr is changed
 path = ListenerPath(
-    node=SimpleTraitListener(name="inst", notify=True),
+    node=RequiredTraitListener(name="inst", notify=True),
     next=ListenerPath(
-        node=SimpleTraitListener(name="attr", notify=True),
+        node=RequiredTraitListener(name="attr", notify=True),
         next=None,
     )
 )
@@ -191,9 +206,9 @@ path = ListenerPath(
 # do not fire if inst is changed (even if it is changed such that
 # inst.attr is changed)
 path = ListenerPath(
-    node=SimpleTraitListener(name="inst", notify=False),
+    node=RequiredTraitListener(name="inst", notify=False),
     next=ListenerPath(
-        node=SimpleTraitListener(name="attr", notify=True),
+        node=RequiredTraitListener(name="attr", notify=True),
         next=None,
     ),
 )
@@ -202,17 +217,17 @@ path = ListenerPath(
 # This is equivalent to having two paths
 paths = [
     ListenerPath(
-        node=SimpleTraitListener(name="inst1", notify=True),
+        node=RequiredTraitListener(name="inst1", notify=True),
         next=ListenerPath(
-            node=SimpleTraitListener(name="attr", notify=True),
+            node=RequiredTraitListener(name="attr", notify=True),
             next=None,
         )
     ),
     # path2
     ListenerPath(
-        node=SimpleTraitListener(name="inst2", notify=True),
+        node=RequiredTraitListener(name="inst2", notify=True),
         next=ListenerPath(
-            node=SimpleTraitListener(name="attr", notify=True),
+            node=RequiredTraitListener(name="attr", notify=True),
             next=None,
         )
     ),
@@ -220,17 +235,17 @@ paths = [
 
 # "inst.container"
 path = ListenerPath(
-    node=SimpleTraitListener(name="inst", notify=True),
+    node=RequiredTraitListener(name="inst", notify=True),
     next=ListenerPath(
-        node=SimpleTraitListener(name="constainer", notify=True),
+        node=RequiredTraitListener(name="constainer", notify=True),
     ),
 )
 
 # "inst.container_items"
 path = ListenerPath(
-    node=SimpleTraitListener(name="inst", notify=True),
+    node=RequiredTraitListener(name="inst", notify=True),
     next=ListenerPath(
-        node=SimpleTraitListener(name="constainer", notify=False),
+        node=RequiredTraitListener(name="constainer", notify=False),
         next=ListenerPath(
             node=ListItemListener(notify=True),
             next=None,
@@ -241,21 +256,21 @@ path = ListenerPath(
 # "inst.[inst1, inst2].attr"
 paths = [
     ListenerPath(
-        node=SimpleTraitListener(name="inst", notify=True),
+        node=RequiredTraitListener(name="inst", notify=True),
         next=ListenerPath(
-            node=SimpleTraitListener(name="inst1", notify=True),
+            node=RequiredTraitListener(name="inst1", notify=True),
             next=ListenerPath(
-                node=SimpleTraitListener(name="attr", notify=True),
+                node=RequiredTraitListener(name="attr", notify=True),
                 next=None,
             )
         ),
     ),
     ListenerPath(
-        node=SimpleTraitListener(name="inst", notify=True),
+        node=RequiredTraitListener(name="inst", notify=True),
         next=ListenerPath(
-            node=SimpleTraitListener(name="inst2", notify=True),
+            node=RequiredTraitListener(name="inst2", notify=True),
             next=ListenerPath(
-                node=SimpleTraitListener(name="attr", notify=True),
+                node=RequiredTraitListener(name="attr", notify=True),
                 next=None,
             ),
         ),
@@ -264,7 +279,7 @@ paths = [
 
 # "inst.attr?"
 path = ListenerPath(
-    node=SimpleTraitListener(name="inst", notify=True),
+    node=RequiredTraitListener(name="inst", notify=True),
     next=ListenerPath(
         node=OptionalTraitListener(name="attr", notify=True),
         next=None,
@@ -273,8 +288,8 @@ path = ListenerPath(
 
 # "inst.attr*"
 # this should match inst.attr.attr.attr.attr....
-attr = SimpleTraitListener(name="attr", notify=True)
-path = ListenerPath(node=SimpleTraitListener(name="inst", notify=True))
+attr = RequiredTraitListener(name="attr", notify=True)
+path = ListenerPath(node=OptionalTraitListener(name="inst", notify=True))
 path.next = path
 
 
@@ -284,7 +299,7 @@ path = ListenerPath(
     next=ListenerPath(
         node=OptionalTraitListener(name="inst1", notify=True),
         next=ListenerPath(
-            node=OptionalTraitListener(name="attr", notify=True),
+            node=RequiredTraitListener(name="attr", notify=True),
             next=None,
         ),
     )
@@ -292,7 +307,7 @@ path = ListenerPath(
 
 # Listen to changes to the items in the values of Dict(Str(), List())
 path = ListenerPath(
-    node=SimpleTraitListener(name="mapping", notify=True),
+    node=RequiredTraitListener(name="mapping", notify=True),
     next=ListenerPath(
         node=DictValueListener(notify=True),
         next=ListenerPath(
@@ -303,7 +318,14 @@ path = ListenerPath(
 
 # Listen to all traits with a metadata 'updated'
 path = ListenerPath(
-    node=AnyTraitListener(filter=lambda trait: "updated" in trait.__dict__),
+    node=FilteredTraitListener(filter=lambda trait: "updated" in trait.__dict__),
     next=None,
 )
 
+
+# Listen to any traits, then any traits of each of these traits...
+path = ListenerPath(
+    node=AnyTraitListener(),
+    next=None,
+)
+path.next = path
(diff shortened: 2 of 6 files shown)


kitchoi:
Updated the main comment to provide more detailed description for discussion. While this is still a work-in-progress (see task lists for what are still missing), it is getting in shape enough for a high-level review for design for those interested/have the time.

kitchoi:
(Update)
Done this task:
> Hooking/rehooking extended listeners for when a list is replaced with a new list that compares equal to the old one

Added bonus: It is possible to switch back to the old behaviour easily.

kitchoi:
Updated: I synchronized the `TraitList` from #912 to here.

poc/trait_list_object.py
```diff
@@ -0,0 +1,895 @@
...
+                start = len(self)
+            if index.stop is not None:
+                stop = self._normalize_index(index.stop)
+            else:
+                stop = 0
+
+        return slice(start, stop, index.step)
+
+    def _should_notify(self, removed, added):
+        return True
```
kitchoi: This is changed from #912 (equivalent to removing all `if _should_notify` in other places).

poc/trait_list_object.py
```diff
@@ -0,0 +1,895 @@
...
+            index : slice
+                The slice(current_length,new_length) if successful if
+                the iterable supports len() else the slice(current_len,None,1).
+            added : list
+                The newly added items.
+            removed : list
+                Will be []
+
+        """
+        iterable = list(iterable)
```
kitchoi: Difference from #912: I added this line (addressing my own comment there).

poc/trait_list_object.py
```diff
@@ -0,0 +1,895 @@
...
+        for notifier in self._notifiers(True):
+            # We don't actually have the old copy of the list
+            notifier(object, self.name, trait_list, event)
+        # -------------------------------------------
+
+    # -------------------------------------------
+    # Added for observer
+    def _notifiers(self, force_create):
+        return self._event_ctrait._notifiers(force_create)
+    # -------------------------------------------
```
kitchoi: This is added on top of what is in #912.

poc/trait_list_object.py
```diff
@@ -0,0 +1,895 @@
...
+    def _get_length(self, object):
+        if isinstance(object, list):
+            return len(object)
+        else:
+            return 1
+
+
+from interfaces import INotifiableObject
+
+INotifiableObject.register(TraitListObject)
```
kitchoi: This is added on top of #912.

poc/trait_types.py
```diff
@@ -0,0 +1,144 @@
...
+        return cls._items_event
+
+    # ---------------------------
+    # Added for observe
+    # ---------------------------
+
+    def as_ctrait(self):
+        trait = super(List, self).as_ctrait()
+        trait.comparison_mode = ComparisonMode.identity
+        return trait
```
kitchoi: This is the only addition compared to one from `traits.trait_types`. Actually if I did the same change on master, no tests fail.

kitchoi:
Edited main comment on:
> ~I created a new subclass NewTraitListObject purely for the POC purpose. It is incomplete but complete enough for testing the listener hooking mechanism.~ Edited: Also synchronized with #912 now

> ~I implemented a new List(TraitType) purely for the POC purposes. It is incomplete but complete enough to test the listener hooking mechanism.~ Edited: It is now synchronized with master.

kitchoi:
Ah... I don't care about flake8 for this POC :(  Switching it off for now

poc/interfaces.py
```diff
@@ -0,0 +1,51 @@
...
+    """ Interface for a notifier.
+    """
+
+    def __call__(self, object, name, old, new):
+        """ Called by ctraits. """
+        raise NotImplementedError()
+
+    def increment(self):
+        """ Increment an internal reference count.
+        Subclass can ignore if reference counting is not required.
```
corranwebster: This is a little smelly... it feels like whether or not reference counting is needed should really be an internal matter for the implementer, but I could be convinced otherwise, because the alternative seems to make the API essentially be `add_notifier`/`remove_notifier` with a lot of duplication between implementations.

poc/observe.py
```diff
@@ -0,0 +1,764 @@
...
+    notifier : INotifier
+    """
+
+    observer_notifiers = object._notifiers(True)
+    logger.debug("Removing from %r", observer_notifiers)
+    for other in observer_notifiers[:]:
+        if other.equals(notifier):
+            other.decrement()
+            if other.can_be_removed():
+                # TODO: Always do this on the main thread!
```
corranwebster: That's _hard_ without substantial changes to Traits internals.  There are some things that you can do to make things more robust:

- guard the `remove` in a `try: ... except ValueError: ...` so the remove operation is idempotent
- similarly, make sure `dispose` is idempotent, and ensure that it is documented as a requirement of the API call

That should cover race conditions where two threads are removing the notifier at the same time.

There's still going to be a race conditions where you have multiple threads adding and removing the same notifier where the adding notifier finds a match, but then before it can increment the counter, the removing thread decrements and tests.  In this case the notifier gets incorrectly removed.  You can protect against this somewhat by having the `increment()` call fail if the refcount is less than or equal to 0 (either throw an exception, or return a success flag) and catch the failure if it happens and fall through to adding a new notifier.  But I don't think it can catch everything.

The alternative is to have every notifier have a `RLock` (or something similar) which has to be acquired while messing around with the refcount.  I worry about the possibility of deadlocks if we were to do that, so maybe it needs to be a single global `RLock`?

I think this all would need some serious thought - particularly with Mark's input, and is probably beyond the scope of this PR.

poc/observe.py
```diff
@@ -0,0 +1,764 @@
...
+        yield from ()
+
+    def get_change_notifier(self, callback, path, target, dispatcher):
+        """ Return a ListenerChangeNotifier for removing/propagating listeners
+        for the 'next' targets.
+        """
+        raise NotImplementedError()
+
+
+class AnyTraitListener(BaseListener):
```
corranwebster: This might need an `optional` flag just like the named trait case. Eg. in the current notation, handling something like `+:foo`.

poc/interfaces.py
kitchoi: That's a good point. Yes I think I can move some bits of the `add_notifier` and `remove_notifier` back into the notifier so to remove this optional interface.

poc/observe.py
mdickinson: > That should cover race conditions where two threads are removing the notifier at the same time.

I don't think this is a case we should worry about. Traits is not thread-safe, and with the exception of a few places that matter, it doesn't try to be thread safe. I don't think there's any particular reason to aim for thread safety in this particular part of the code.

poc/observe.py
```diff
@@ -0,0 +1,764 @@
...
+            return True
+        if type(self) is not type(self):
+            return False
+        return (
+            (self.metadata_name, self.include)
+            == (other.metadata_name, other.include)
+        )
+
+
+def MetadataTraitListener(metadata_name, notify, include):
```
corranwebster: I think that there might be a better way of implementing this independently of `_FilteredTraitListener` which is more natural for metadata (the `_FilteredTraitListener` class may be independently useful, but I'd make it part of the API).  I'd have the API look something like:
```
from traits.trait_base import not_none

class MetadataTraitListener(BaseListener):
    def __init__(self, metadata_name, notify, filter=not_none):
        ...

   def iter_this_targets(self, object):
        # object must be an instance of HasTraits
        for name in object.trait_name(self.metdata_name=self.filter).items():
            yield object._trait(name, 2)

    ... # etc
```
This is more in line with the way that expert users are used to filtering traits.

poc/observe.py
corranwebster: Perhaps we should take the opportunity to remove support for `dispatch='new'` in the new interface?

mdickinson: Sounds good to me; it can always be re-added when someone demonstrates that there's a real need for it.

kitchoi:
Update: 
`trait_added` is now handled by thees changes (the last 8 commits):
https://github.com/enthought/traits/pull/942/files/62e2092b851c2faa3acfc43f0ad682718fd4ed10..3839ccdf098afb8d7a278dad56f8e51b51e59f9e

kitchoi:
Update: Just reset the `TraitListObject` back to the copy from master, and added a few necessary changes.
Now all the tests should pass... let's see what CI says.

poc/observe.py
```diff
@@ -0,0 +1,865 @@
...
+            return True
+        if len(self.nexts) != len(other.nexts):
+            return False
+        for nexts in itertools.permutations(other.nexts, len(other.nexts)):
+            if all(n1 == n2 for n1, n2 in zip(self.nexts, nexts)):
+                return True
+        return False
+
+    @classmethod
+    def from_nodes(cls, node, *nodes):
```
kitchoi: See #969 for constructing `ListenerPath` in a more user-friendly manner. This method might be useful to be written in the tests though.

kitchoi:
I only just came across #529 and I am going to test (locally) whether this listener framework will allow us to fix the issue.

kitchoi:
> I only just came across #529 and I am going to test (locally) whether this listener framework will allow us to fix the issue.

The solution is not obvious, but is no less obvious than the original framework. Popping it off my task list for now.

mdickinson:
@kitchoi I'm having difficulty understanding the big picture here. Is there any documentation or anything I can look at that sketches out how things fit together.

One specific question: `INotifier.__call__` has signature `(self, arg1, arg2, arg3, arg4)`, but the `ctraits.c` internals always expect a signature of `(obj, name, old, new)`.

Is the generic `(arg1, arg2, arg3, arg4)` because we may be using other values for that `__call__`? Or can those arguments be renamed to `(obj, name, old, new)`?

kitchoi:
@mdickinson Indeed this `(object, name, old, new)` to `(arg1, arg2, arg3, arg4)` is a late generalization I made. In fact, it is the last commit: 7ff38a90deb5f7a76eae4c35a8db79f6836f6846
I also omitted to comment on the rationale for this update, I am sorry about that.

This change is motivated by some awkwardness I encountered:
When a nested `TraitList` is mutated, the `object`, `name`, and `old` are not well defined. If we fill some values for `object`, `name` and `old`, these values would be very confusing/misleading.

I also had difficulties reusing the `TraitList.notify(trait_list, index, removed, added)` interface because the call signature does not match `object, name, old, new`.

This is when I realized,
(1) the `BaseObserverEvent` can only guarantee that `new` is meaningfully defined. 
(2) Listeners being context aware (e.g. `NamedTraitListener` expects notifications from `ctrait.c`, and `ListItemListener` expects notifications from `TraitList.notify`), means that they can impose meaning on the four arguments in their corresponding `event_factory`.

As soon as I strip the meaning from `(object, name, old, new)` from the notifier call signature, I also moved the `if old is not Uninitialized` to the listeners' event factory.

mdickinson:
Okay, so this isn't just about naming. In that case, I don't think `INotifier` makes sense as an interface: we really have two different interfaces at that point.

kitchoi:
The `BaseListener` could be called an interface too? Or not call `INotifier` an interface or any of these an "interface"?

poc/observe.py
```diff
@@ -0,0 +1,865 @@
...
+            return True
+        if type(other) is not type(self):
+            return False
+        if self.node != other.node:
+            return False
+        if self.nexts == other.nexts:
+            return True
+        if len(self.nexts) != len(other.nexts):
+            return False
+        for nexts in itertools.permutations(other.nexts, len(other.nexts)):
```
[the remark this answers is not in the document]
kitchoi: Any suggestion around the performance concern?

mdickinson:
There may be a useful interface lying underneath the `BaseListener`, though that interface I think wouldn't include `event_factory`, since that method isn't useful without knowing more about the type of the concrete listener.

Similarly, there may be something useful in `INotifier`, but you wouldn't want to include the `__call__` method in that interface, since you can't use it without knowing more about the object.

For an interface to be useful, it needs to be possible to do things with an object even if all you know about that object is that it satisfies the interface; at the moment, I'm not seeing where we can make use of `INotifier` that way. But I'm still getting my head around the big picture, and I'm still missing lots of things.

Some background documentation that sets out what the key bits are here and how they work together would be really useful.

poc/trait_observer_notifier.py
```diff
@@ -0,0 +1,193 @@
...
+import weakref
+
+from traits.trait_base import Uninitialized
+
+logger = logging.getLogger()
+
+
+class BaseObserverEvent(object):
+
+    def __init__(self, arg1, arg2, arg3, arg4):
```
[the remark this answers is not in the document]
kitchoi: Agreed the `__init__` can be removed here.

I wanted to note that an observer event has only one guarantee attribute: `new`. Everything else is context dependent, e.g. list change event does not provide `object`, `name`, `old`.

poc/observe.py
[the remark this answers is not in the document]
mdickinson: I haven't looked at the details. Can we make things hashable, and compare sets of objects?

poc/interfaces.py
```diff
@@ -0,0 +1,51 @@
...
+    notifiers can be added / removed.
+    """
+
+    @abc.abstractmethod
+    def _notifiers(self, force_create):
+        return []
+
+
+class INotifier(abc.ABC):
+    """ Interface for a notifier.
```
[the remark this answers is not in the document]
kitchoi: At the end I did not declare any `abstractmethod` and its friends. So I can remove the `abc.ABC` here.

poc/interfaces.py
```diff
@@ -0,0 +1,51 @@
...
+        raise NotImplementedError()
+
+    def can_be_removed(self):
+        """ Return true if this notifier can be removed from a
+        notifiable object.
+        """
+        # return true or false.
+        raise NotImplementedError()
+
+    def dispose(self):
```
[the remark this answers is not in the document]
kitchoi: That is the part I inherited from #931
It seems that it is motivated by the existing `TraitChangeNotifier` which releases any reference to `object`.

poc/trait_observer_notifier.py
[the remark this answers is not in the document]
mdickinson: Are there contexts where all we care about or use is the `new` attribute? That is, is there value in having an interface or abstract base class that encapsulates that?  For example, is there code in this PR that deals with both types of events but only makes use of the `new` attribute?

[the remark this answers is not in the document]
kitchoi: The `new` is not used by the listener mechanism... but I guess this is something a user wants to know when they do something like this:
```
@observe(...)
def _handle_change(self, event):
    ...
```

Alternatively, we can remove `new` and then document separately what the user will receive based on the type of events (where `new` appears in all of them I guess).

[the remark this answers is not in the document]
mdickinson: Right, the user presumably knows from context what type of event they're receiving (and has to know what type of event they're receiving to be able to do anything useful with it). I think that kills the idea of a useful interface here.

kitchoi:
Updated with three commits:
https://github.com/enthought/traits/pull/942/files/7ff38a90deb5f7a76eae4c35a8db79f6836f6846..ab4833de48736d7db8cdcaac29c159d6c6c7a02d

The second one is an orthogonal change: It reuses the `TraitListEvent`. But I unwrap it again in the user-facing `ListObserverEvent` because "Flat is better than nested".

The first and third commits changes the `INotifier.__call__` signature to `*args`.  Both `ListenerChangeNotifier` (for maintaining listeners) and `TraitObserverNotifier` (for calling user's callback) are changed. And then four subclasses are defined to redefine the signature of `__call__` again.  There are four because this POC currently handles notifiers from CHasTraits and `TraitListObject` only.  If we had included `TraitDictObject` and `TraitSetObject`, there will be eight subclasses.

In all of these notifiers, the arguments are sent straight through to an event factory that adapts the arguments into an object.

poc/trait_observer_notifier.py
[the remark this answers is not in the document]
kitchoi: `BaseObserverEvent` is now gone: https://github.com/enthought/traits/pull/942/commits/205004fe3a67f99088bcdb34241d8d5df861f87e

poc/interfaces.py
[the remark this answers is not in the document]
kitchoi: `abc.ABC` is removed.

kitchoi:
Update: `add_notifier` and `remove_notifier` are moved inside the implementations of `INotifier` because reference counting is only implemented by `TraitObserverNotifier` and not `ListenerChangeNotifier`.

kitchoi:
> Updated with three commits:
https://github.com/enthought/traits/pull/942/files/7ff38a90deb5f7a76eae4c35a8db79f6836f6846..ab4833de48736d7db8cdcaac29c159d6c6c7a02d
> The first and third commits changes the INotifier.__call__ signature to *args. Both ListenerChangeNotifier (for maintaining listeners) and TraitObserverNotifier (for calling user's callback) are changed. And then four subclasses are defined to redefine the signature of __call__ again. There are four because this POC currently handles notifiers from CHasTraits and TraitListObject only. If we had included TraitDictObject and TraitSetObject, there will be eight subclasses.

The subclassing made the `event_factory` redundant, and the event factory is removed in 5432c7fe78f23eca92b3aae5f7070223fa318718
But as a proof-of-concept, I tested how listeners can be extended for filtering events, e.g. on comparison mode:
https://github.com/enthought/traits/blob/7ff4641e47981554abf0b30844a66de1fe5d8a0b/poc/experiments.py#L558-L597
For this I added `prevent_event` as a callable, to replace `event_factory` returning None.
Note that this particular feature may not be needed afterall, so it probably won't be included for production. It is just a test on how such a feature could be implemented.

Note also this subclassing pattern is quite similar to the existing many subclasses of `TraitChangeNotifierWrapper` (in `trait_notifiers.py` on master), where subclasses are created for implementing different `dispatch`. Here `dispatch` has been made a separate callable as an input parameter, so we don't have to subclass for call signatures AND subclass for different dispatch.

kitchoi:
Since I failed to use one commit to go from event_factory to subclassing and it is hard to read across non-consecutive commits, I created one for going the opposite direction: From subclassing to event_factory. https://github.com/enthought/traits/pull/942/commits/45559d5c62870345602d9c4d3d8aef7f0a4af6ed

This commit should make it easier to compare the two approaches. Pinging @mdickinson for this was what we discussed yesterday.

poc/observe.py
```diff
@@ -0,0 +1,689 @@
...
+    "same": dispatch_same,
+    "new": dispatch_new_thread,
+}
+
+
+def observe(object, callback, path, remove, dispatch):
+    """
+    Parameters
+    ----------
+    object : HasTrait
```
mdickinson, line 60: Question: the type (`HasTraits`?) doesn't quite seem to match the description here. Which one is right?

poc/observe.py
```diff
@@ -0,0 +1,689 @@
...
+        An object that implements `_notifiers` for returning a list for
+        adding or removing notifiers
+    callback : callable(object)
```
mdickinson, line 63: Is there a better name we can use here? `callback` is a bit generic.

poc/observe.py
```diff
@@ -0,0 +1,689 @@
...
+
+# These are user-facing notifier factories...
+DISPATCHERS = {
+    "extended": dispatch_same,
+    "same": dispatch_same,
+    "new": dispatch_new_thread,
+}
+
+
+def observe(object, callback, path, remove, dispatch):
```
mdickinson, line 56: I think we want to make `remove` and `dispatch` keyword-only here (and perhaps some of the other arguments, too).

Could we also shuffle the signature a bit to place `object` and `path` together? IIUC, those two arguments comprise the "what" that we're observing

poc/observe.py
```diff
@@ -0,0 +1,689 @@
...
+        add_notifiers(
+            path=path,
+            object=object,
+            callback=callback,
+            target=object,
+            dispatcher=DISPATCHERS[dispatch],
+        )
+
+
+def add_notifiers(object, callback, path, target, dispatcher):
```
mdickinson, line 95: Could you add `target` to the docstring? What's the difference in roles of `object` and `target`?

poc/observe.py
kitchoi: How about `handler`? `on_trait_change` calls this `handler`.

poc/observe.py
```diff
@@ -0,0 +1,689 @@
...
+    # what is going to happen in TraitDictObject
+
+    def __init__(self, notify):
+        self.notify = notify
+
+
+class ListenerPath:
+
+    def __init__(self, node, nexts=()):
+        self.node = node
```
mdickinson, line 653: What are the types of `node` and `nexts` here, and what do they represent?

poc/observe.py
kitchoi: Yes. I had a (hopefully-better) description in `ListenerChangeNotifier`, which I can copy over here:
```
        target : Any
            Sets the context for the (actual) callback. In practice, it is
            the HasTrait instance on which the listener path is defined and
            will be seen by users as the "owner" of the notifier.
            Strictly speaking, this object sets the context for the notifier
            and does not have to be a notifiable object.
```

kitchoi: > What's the difference in roles of object and target?

`object` is where the notifiers are going to be attached to. `target` is the context.
In `observer`, `object` and `target` are the same, but as we walk down the path to nested objects, `object` is the nested object onto which notifiers are attached, whereas `target` is kept unchanged (the mother object if this is a better name?).

kitchoi: > In observer,

Sorry, in `poc.observe.observe`, not `observer`.

poc/observe.py
kitchoi: `node` is an instance of `BaseListener` (described in the main comment).
I removed the examples a few commits ago considering those as noise, but `nexts` are list of `ListenerPath`.

kitchoi: Examples were here, not sure if they help or not:
https://github.com/enthought/traits/pull/942/commits/7ff4641e47981554abf0b30844a66de1fe5d8a0b

poc/observe.py
mdickinson: Right, `handler` sounds a bit better. Maybe `listener`?

poc/observe.py
mdickinson: Still confused by the description here. `HasTraits` *does* implement `_notifiers`, but that's not something that we usually use (it's used for static `anytrait` listeners.) Does the new observe functionality depend on `HasTraits._notifiers` somewhere?

kitchoi: Ah right. The description is bogus, I am sorry for the confusion.

Maybe this is better, or something along the line of:
```
object : HasTraits
   An instance on which trait changes are observed according to the given attribute path.
```
Other suggestions are welcome too.

> Does the new observe functionality depend on HasTraits._notifiers somewhere?

No the new observe functionality does not depend on `HasTraits._notifiers`.
Previously I had an instance of `BaseListener`  for `anytrait`, but I removed it because we decided to ditch anytrait: https://github.com/enthought/traits/pull/942/commits/9b37f6785578054877d42e233501667324625455
Note that the `iter_this_target` simply yield the given object, which is assumed to be an instance of `HasTraits` in that context.

poc/observe.py
kitchoi: I think I would want a word that sounds more active, so I prefer `handler` to `listener`. naming-thing-is-hard...

poc/observe.py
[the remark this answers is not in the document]
kitchoi: Good idea!  I did this in #976

poc/interfaces.py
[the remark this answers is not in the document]
kitchoi: The `remove_notifier` is now moved to the `INotifier` as `remove_from`. So this `dispose` can be removed from the interface. Subclass will do their cleanup in `remove_from`.

poc/observe.py
```diff
@@ -0,0 +1,701 @@
...
+
+            for next_target in listener.iter_next_targets(object):
+                add_notifiers(
+                    object=next_target,
+                    callback=callback,
+                    path=next_path,
+                    target=target,
+                    dispatcher=dispatcher,
+                )
+    if type(listener) is not TraitAddedListener:
```
mdickinson, line 140: Is there a way that this could be absorbed into the `BaseListener` "interface"? This explicit type check seems like a form of OCP violation.

kitchoi: Yes!

kitchoi: Note: This check is to prevent recursion, because it calls `add_notifiers` again with `TraitAddedListener` as the first listener.

kitchoi: Done: https://github.com/enthought/traits/pull/942/commits/7e87fb35476180a042b9b525c42ef42ffb0490ea

Now the recursion is avoided by the `TraitAddedListener.get_static_paths` not returning any paths that have `TraitAddedListener` as root (it returns an empty list at the moment).

kitchoi:
Update:
This commit extends support for `TraitDictObject`: fcee637187895cb0a63aea41c94884cb43dd0916
This commit integrates changes with the new `TraitListObject` recently merged to master: f6a0738f1e573b1515e8d97f6cf3526c3ef003ad

kitchoi:
Given this branch has now served the purpose of proving a concept and illicit discussion, I am going to close this PR and redirect future discussion to issue #977.  I have also added a README.rst file to describe relationships between objects in this branch. Note that the naming in this branch may be changed. For a start, all "listeners" will be renamed to "observers".

Note that there are a few comments not addressed in this PRs, e.g. https://github.com/enthought/traits/pull/942#discussion_r401481463
When we implement the code for production, we should revisit those comments.

PR #951, opened 2020-03-20
An approach to silence events from nested containers.

midhun-pm:
**Checklist**
- [x] Tests

Opened this PR for discussion/suggestions on the approach.(Description will be improved)

The objective is to somehow figure out which container is the outermost one and enable it to fire events.

All its children are not allowed to fire notifications.

Approach Outline:
Since the innermost containers are initialized first, it is initially assumed to be a root and marked accordingly, until its parent is created and we discover that our assumption is wrong un-marking it as a root. This process happens until we discover the correct root.

Before firing an event, all we need to do now is check if it is marked a root and fire only if it is one.

### DIFF
--- a/traits/trait_list_object.py
+++ b/traits/trait_list_object.py
@@ -79,6 +79,8 @@ class TraitListObject(list):
         The name of the attribute on the object.
     value : list
         The list of values to initialize the TraitListObject with.
+    is_root_container : bool
+        True if the container has no children.
 
     Attributes
     ----------
@@ -100,10 +102,9 @@ class TraitListObject(list):
         self.object = ref(object)
         self.name = name
         self.name_items = None
+        self.is_root_container = is_root_container
         if trait.has_items:
             self.name_items = name + "_items"
-
-        self.is_root_container = is_root_container
 
         # Do the validated 'setslice' assignment without raising an
         # 'items_changed' event:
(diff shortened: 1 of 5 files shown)


midhun-pm:
The current code will not fix the bug for bare trait declarations. i.e. For example: `List(Dict)` will not work, but changing it to `List(Dict())` will.  
Fix works for `List` `Dict` `Set` `Union` `Either` and combinations of these.

traits/trait_types.py
```diff
@@ -2366,7 +2385,8 @@ def validate(self, object, name, value):
             if object is None:
                 return value
 
-            return TraitListObject(self, object, name, value)
+            return TraitListObject(self, object, name, value,
+                                   self.is_root_container)
```
midhun-pm: Need to check if `is_root_container` attribute exists

traits/trait_list_object.py
```diff
@@ -124,6 +127,9 @@ def __init__(self, trait, object, name, value):
     def _send_trait_items_event(self, name, event, items_event=None):
         """ Send a TraitListEvent to the owning object if there is one.
         """
+        if not self.is_root_container:
+            return
+
```
kitchoi: Have you considered this:
```
object = self.object()
if not getattr(object, self.name) is self:
    return
```

midhun-pm: This is good! Tried it, works if the root is a `List` or `Dict` or `Set`, but doesn't if the container was inside something like `Either` or `Union` `Either(Set(), None)`.

midhun-pm:
Closing this PR, will be fixed once EEP-3 is in.

PR #995, opened 2020-04-08
Simplify checks in TraitDictObject.notifier and add the related test

kitchoi:
This PR is targeting #913 (TraitDictObject) and is similar to #990 which was targeting TraitSetObject.

This PR simplifies the checks in `TraitDictObject.notifier` and add a test that would fail if we did not check for `if self.name_items is None`.

### DIFF
--- a/traits/tests/test_dict.py
+++ b/traits/tests/test_dict.py
@@ -11,6 +11,7 @@
 """ Test cases for dictionary (Dict) traits. """
 
 import unittest
+from unittest import mock
 
 from traits.trait_types import Any, Dict, Event, Str, TraitDictObject
 from traits.has_traits import HasTraits, on_trait_change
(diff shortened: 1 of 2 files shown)


kitchoi:
Note that there is a test in #913 that would fail on Python 3.5 due to dictionary random ordering. I did not fix it here and the CI result may be affected by that.

traits/tests/test_dict.py
```diff
@@ -146,3 +147,21 @@ class Foo(HasTraits):
...
             foo.mapping["a"] = 1
+
+    def test_items_set_to_false(self):
+
+        class Foo(HasTraits):
+
+            mapping = Dict(items=False)
+
+        handler = mock.Mock()
+        # Setting items to false effectively switches off
```
midhun-pm, line 158: Not sure why we have this ability.

kitchoi: Neither do I... but it is there :/

midhun-pm:
LGTM

PR #1007, opened 2020-04-15
Add notifier for maintaining nested notifiers in the observer framework

kitchoi:
This PR add `ObserverChangeNotifier` for maintaining notifiers in nested objects in the observer framework. It used to be called `ListenerChangeNotifier` in #942

Contribute towards #977

See #1000 also for the `TraitEventNotifier` that wraps the user handler. This `ObserverChangeNotifier` is used for removing/adding the nested `TraitEventNotifier` when a container object is replaced. The test in `TestIntegrationHasTraits` gives a rough idea how this works.

Edited: #1023 may provide more context. The `ObserverChangeNotifier` is typically provided for the `get_maintainer` method in `IObserver`.

**Checklist**
- [x] Tests
- ~Update API reference (`docs/source/traits_api_reference`)~: Nothing here is intended to be exposed publicly.
- ~Update User manual (`docs/source/traits_user_manual`)~: Nothing here is intended to be exposed publicly.
- ~Update type annotation hints in `traits-stubs`~: Nothing is trait-ed here.


traits/observers/_observer_change_notifier.py
```diff
@@ -0,0 +1,167 @@
...
+            notifier will be silenced.
+        target : object
+            An object for defining the context of the user's handler notifier.
+            A weak reference is created for the target. If the target is
+            garbage collected, this notifier will be silenced.
+            This is typically an instance of HasTraits acting as a container
+            for an observable on which this notifier is attached to. It would
+            be seen by the user as the "owner" of the observer.
+        dispatcher : callable(function, event)
+            Callable for dispatching the user's handler.
```
kitchoi, line 64: I think these `handler, target, dispatcher` could do being wrapped in another object as they do go together. See #1000. I can do that after both PRs are in.

traits/observers/_observer_change_notifier.py
```diff
@@ -0,0 +1,167 @@
...
+            and self.observer_handler is other.observer_handler
+            # path is an input for observer_handler.
+            # Unequal paths should not interfere each other.
+            and self.path == other.path
+            # user handler is an input for observer_handler.
+            # different user handlers should not interfere each other.
+            and self.handler() is other.handler()
+            # target is an input for observer_handler.
+            # it goes together with the user's handler
+            and self.target() is other.target()
```
kitchoi, line 162: Same question as in #1000, should we differentiate notifiers with different dispatch?  The existing on_trait_change does not, but here is an opportunity to change that.

kitchoi: It is difficult to see why these checks are needed out of the larger, integrated context.
I added a few tests on the PoC branch to convince myself that these checks are necessary: https://github.com/enthought/traits/compare/f34fa3920d41ec6cbe9d513cefcc35fd5c7edc65...986fcd25fc68f867d4e4421eba90dc7619231033

kitchoi: In my test branch where I have cleaned up things, the last test shown in the diff does not fail when `self.target() is other.target()` is removed any more. 
But this one does (using the `observer-framework-test` branch). The last assertion would have failed with `1 != 0` if targets are not compared.

<details>
<summary> A test that justifies checking target identity </summary>

```

class Potato(HasTraits):

    name = Str()


class PotatoBag(HasTraits):

    potatos = List(Instance(Potato))


class Crate(HasTraits):

    potato_bags = List(PotatoBag)


class TestHasTraitsObserverDifferentiateParent(unittest.TestCase):

    def test_shared_instance_same_graph_different_target(self):

        crate1 = Crate()
        crate2 = Crate()

        # given
        handler = mock.Mock()
        crate1.observe(handler, "potato_bags:items:potatos:items:name")
        crate2.observe(handler, "potato_bags:items:potatos:items:name")

        new_potato = Potato()
        new_potato_bag = PotatoBag(potatos=[new_potato])
        crate1.potato_bags = [new_potato_bag]
        crate2.potato_bags = [new_potato_bag]
        new_potato.name = "King Edward I"
        self.assertEqual(handler.call_count, 2)
        handler.reset_mock()

        # when
        # remove the second observer
        crate2.observe(
            handler, "potato_bags:items:potatos:items:name", remove=True)
        new_potato.name = "King Edward II"

        # then
        self.assertEqual(handler.call_count, 1)
        handler.reset_mock()

        # then
        # This check the observer is maintained.
        maris_piper = Potato()
        crate2.potato_bags[0].potatos.append(maris_piper)
        crate1.potato_bags = []
        self.assertEqual(handler.call_count, 0)  # sanity check

        maris_piper.name = "Maris Piper"
        self.assertEqual(handler.call_count, 0)
```

</details>

traits/observers/_observer_change_notifier.py
kitchoi: Resolving... I will do this refactoring later when all the pieces are here. This change is not going to affect the user API.  (see discussion in #1000)

traits/observers/_observer_change_notifier.py
kitchoi: > Same question as in #1000, should we differentiate notifiers with different dispatch? The existing on_trait_change does not, but here is an opportunity to change that.

I am going to add the comparison on the `dispatcher` here... see discussion in #1000.

traits/observers/_observer_change_notifier.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        self.observer_handler = observer_handler
+        self.event_factory = event_factory
+        self.prevent_event = prevent_event
+        self.graph = graph
+        self.target = weakref.ref(target)
+        if isinstance(handler, types.MethodType):
+            self.handler = weakref.WeakMethod(handler)
+        else:
+            self.handler = partial(_return, value=handler)
```
corranwebster, line 88: Perhaps a name better than `_return` which helps explain what's going on, or just use a lambda.  I had to think about what was happening and go searching around for `_return`.

traits/observers/_observer_change_notifier.py
```diff
@@ -0,0 +1,183 @@
...
+            type(self) is type(other)
+            # observer_handler contains the logic for maintaining notifiers
+            # in the downstream graph.
+            and self.observer_handler is other.observer_handler
+            # graph is an input for observer_handler.
+            # Unequal graphs should not interfere each other.
+            and self.graph == other.graph
+            # user handler is an input for observer_handler.
+            # different user handlers should not interfere each other.
+            and self.handler() == other.handler()
```
corranwebster, line 172: This has the possibility of spurious equality in the case where:

- everything is the same _except_ the handler
- you have two different method handlers
- which have both been garbage collected

In this case both will return `None` and compare equal.  Of course in this situation they could both be `None` because it was the _same_ handler which had been gc'd.

You should try to do something sensible if `handler()` returns None, which might be to raise, or it may be something else.  It may even be that this is OK as-is.

kitchoi: If both `self.handler()` and `other.handler()` are initially referring to different instance methods and the instances have been garbage collected, and everything else is the same, then the two notifiers are genuinely indistinguishable from one and other. These notifiers are silenced too.
I think overall not dealing with `handler() is None` is okay...I can't think of a test that would fail where the test represents some actual use case.

traits/observers/_observer_change_notifier.py
kitchoi: I define the function here instead and I will do the same for the other notifier.

PR #1008, opened 2020-04-15
Raise warning when overriding HasTraits methods in subclasses

midhun-pm:
**Checklist**
- [x] Tests
~~- [ ] Update API reference (`docs/source/traits_api_reference`)~~

Part of #705 

PR adds a check for user defined subclass of `HasTraits` and warns if any of the attribute names override public methods in `HasTraits`


midhun-pm:
I'm not sure if these are too many warnings.

traits/has_traits.py
```diff
@@ -115,6 +115,30 @@ class AbstractViewElement(abc.ABC):
...
+    "Interface",
+    "ISerializable",
+    "traits_super"
+)
+
+
+# A custom warning issued by Traits
+class TraitsWarning(Warning):
+    pass
+
```
kitchoi, line 141: I think we would want to expose this class on `api.py` so that downstream projects can import it for capturing the said warning.
Please add a class docstring to it as well.  We will also need to update the documentation to mention this class.

traits/tests/test_warnings.py
```diff
@@ -0,0 +1,34 @@
...
+        for method_name in method_names:
+            with self.subTest(method_name):
+                if not method_name.startswith("_"):
+                    # Ensure that overriding public method
+                    # names raises a warning.
+                    with self.assertWarns(TraitsWarning):
+                        exec(class_def.format(method_name))
+                else:
+                    # But overriding private method names is allowed.
+                    exec(class_def.format(method_name))
```
kitchoi, line 34: If we change to check namespace instead, then we won't need to dynamically compose code in this test, something like this would do too:
```
with self.assertWarns(TraitsWarning):
    class UserDefinedClass(HasTraits):
        trait = "some_value"
```

traits/has_traits.py
```diff
@@ -396,6 +427,13 @@ def update_traits_class_dict(class_name, bases, class_dict):
     # Move all trait definitions from the class dictionary to the
     # appropriate trait class dictionaries:
     for name, value in list(class_dict.items()):
+
+        # Warn name collisions.
+        if check_name_collision and name in public_method_names:
```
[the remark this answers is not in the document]
midhun-pm: I'm in favor of reserving names starting with `traits` or `_traits`. But do we want to do that in addition to or instead of what's being done in this PR ?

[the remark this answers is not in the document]
midhun-pm: What if `HasTraits` introduces a new public method that does not being with `traits` ?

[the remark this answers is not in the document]
mdickinson: > What if `HasTraits` introduces a new public method that does not being with `traits` ?

So the idea of #705 is that Traits would promise not to introduce such new methods: it would use names beginning with `trait` for new public methods, and names beginning with `_trait` for new private methods.

That's exactly the problem that #705 aims to solve: every time we introduce a new method on `HasTraits` we risk conflicting with user-defined trait names. But if we only ever introduce new methods in the `trait*` / `_trait*` namespace, and we tell _users_ of Traits not to name their own traits that way, we avoid that conflict.

traits/has_traits.py
midhun-pm: Is this the right place to define it ? There is a `trait_errors` module for errors but nothing equivalent for warnings . Not sure if I should create a new module, or maybe put it in `trait_errors`?

mdickinson: Why not use one of the built-in warning classes rather than defining our own? The warnings module provides good functionality for filtering by module or by message, so I'm not sure we need a dedicated `Warning` subclass here.

mdickinson: Specifically, I think `UserWarning` would work here. Alternatively, we might consider `DeprecationWarning`: if we go the route of carving off a namespace for Traits' sole use, then we'll eventually want to make used-defined trait names starting with `_trait` or `trait` an error. The only reason not to raise right now is backwards compatibility.

I think the problem with `TraitsWarning` is that it's not clear what it's actually for. If the plan is to use it for any warning issued by Traits now or in the future, then I think that's a bad idea: I think we want our warning type to be related to the type of condition that's being warned about, rather than related to a particular package - we already have module-level filtering for that.

If the plan is to use it for this _specific_ name clash issue, then it probably needs a more descriptive name. And then we risk ending up with various warnings in the future. It seems simpler all round to use the built-in warning classes unless we have a good reason not to.



I see a problem with `TraitsWarning`

I also think the `TraitError` / `TraitsWarning` name mismatch is going to be confusing.

traits/has_traits.py
```diff
@@ -115,6 +115,28 @@ class AbstractViewElement(abc.ABC):
...
+class TraitsWarning(Warning):
+    """ A warning from Traits which could indicate that something
+    in the user code is using traits in a way that is was not
+    designed to be used.
+    """
+    pass
+
+
+def _is_disallowed_prefix(name):
+    for prefix in DisallowedNamePrefixes:
```
mdickinson, line 135: No need for a loop here: you can do `return name.startswith(DisallowedNamePrefixes)`.

traits/has_traits.py
kitchoi: Agreed that the good old `UserWarning` would also do the job here.

traits/has_traits.py
```diff
@@ -374,6 +396,10 @@ def update_traits_class_dict(class_name, bases, class_dict):
     class_dict : dict
         A dictionary of class members.
     """
+    # Enable name collision check for user defined classes.
+    check_name_collision = (False if class_dict.get("__module__", "")
+                            .startswith("traits.") else True)
```
kitchoi, line 401: Nit: It would be better to define this `"traits"` package name as a module level constant. Preferably computed from `__name__`. While it is unlikely that the package will change name, it is rather easy to avoid having to make that assumption.

Nit: The fact that we are doing `startswith("traits.")` instead of `split(".", 1)` here means that if the module name is `"traits"` exactly (i.e. without the `"."`), then `check_name_collision` is true. In other words we will be checking for name collisions for any classes defined in `traits/__init__.py`. There are no classes defined there now, and will unlikely have one... but again, another assumption we can avoid making rather cheaply by using `split`.

traits/has_traits.py
```diff
@@ -396,6 +422,13 @@ def update_traits_class_dict(class_name, bases, class_dict):
     # Move all trait definitions from the class dictionary to the
     # appropriate trait class dictionaries:
     for name, value in list(class_dict.items()):
+
+        # Warn name collisions.
+        if check_name_collision and _is_disallowed_prefix(name):
+            warnings.warn("The attribute named '{}' of class {} uses a "
+                          "prefix that is discouraged. Consider renaming it."
+                          .format(name, class_name), TraitsWarning, 3)
```
kitchoi, line 430: The users will likely want to know exactly what prefixes should not be used are so they don't have to fix this by trial-and-error. Please mention the protected prefixes in the warning message.

traits/has_traits.py
```diff
@@ -2518,6 +2551,10 @@ def add_trait(self, name, *trait):
             it is equivalent to passing the entire list of values to Trait().
 
         """
+        if _is_disallowed_prefix(name):
+            warnings.warn("The attribute named '{}' uses a prefix that is "
+                          "discouraged. Consider renaming it."
+                          .format(name), TraitsWarning, 3)
```
kitchoi, line 2557: Same here, please mention the protected prefixes in the warning message.

traits/tests/test_warnings.py
```diff
@@ -0,0 +1,40 @@
...
+
+from traits.api import Any, HasTraits
+from traits.has_traits import TraitsWarning
+
+
+class TestTraitWarning(unittest.TestCase):
+
+    def test_invalid_trait_name_prefix(self):
+        with self.assertWarns(TraitsWarning):
+            class UserDefinedClass(HasTraits):
```
kitchoi, line 21: Since this class and ones below are also defined in `traits`, the warning is not raised.
The following solution is a bit ugly, but I can't think of a better way apart from mangling the code for tests (which I think is worse):
```

from unittest import mock

def mock_module_name(name):
    return mock.patch(__name__ + ".__name__", name)

class TestTraitWarning(unittest.TestCase):

        with self.assertWarns(TraitsWarning):
            with mock_module_name("somepackage"):
                class UserDefinedClass(HasTraits):
                    trait = "something"
```

traits/has_traits.py
midhun-pm: My thinking was that it would be easier to filter out `TraitsWarning` especially when, for example ,analyzing a log file of some long running application, but I guess the module level filtering covers it and the warning should be more specific. Anyway, I understand why it's better to not have it. Removed.

midhun-pm:
Warnings are raised for `traitsui` , looks like that are attributes in there with prefix `trait`. Maybe I should have a list of excluded modules like `traits` , `traitsui`, etc?

kitchoi:
> Warnings are raised for traitsui , looks like that are attributes in there with prefix trait. 

Interesting! We should look into these warnings.

>  Maybe I should have a list of excluded modules like traits , traitsui, etc?

Hmmm, I would think not, because *ideally* traits should not need to know about traitsui. The warnings might indicate legitimate issues that traitsui needs to fix. In places where the warning is not appropriate or if it is getting in the way of releases, suppressing the warnings **in traitsui** is also an option.

traits/has_traits.py
```diff
@@ -396,6 +419,16 @@ def update_traits_class_dict(class_name, bases, class_dict):
...
     for name, value in list(class_dict.items()):
+
+        # Warn name collisions.
+        if check_name_collision:
+            prefix = _disallowed_prefix(name)
+            if prefix:
+                warnings.warn("The attribute named '{}' of class {} uses a "
+                              "prefix '{}' that is discouraged. Consider "
+                              "renaming it." .format(name, class_name, prefix),
+                              UserWarning, 3)
```
kitchoi, line 430: I am sorry for the misunderstanding: I did not mean to show *the matched prefix*. I was hoping to see all the protected prefixes, namely, `"trait"` and `"_trait"`.  This would save you having to loop over the disallowed prefixes as well.

What you had before was closer...something like:
```
if check_name_collision and name.startswith(DisallowedNamePrefixes):
    warnings.warn("The attribute named '{}' of class {} uses one of these prefixes: {}. Consider renaming it.".format(
name, class_name, ", ".join(DisallowedNamePrefixes))
```

Just mentioning here for future consideration: Consider checking `if prefix is None` instead of just `if prefix`, the former is more explicit and would not be confused with an empty string (which would likely be a programming error).

ievacerny:
> > Warnings are raised for traitsui , looks like that are attributes in there with prefix trait.
> 
> Interesting! We should look into these warnings.
> 
> > Maybe I should have a list of excluded modules like traits , traitsui, etc?
> 
> Hmmm, I would think not, because _ideally_ traits should not need to know about traitsui. The warnings might indicate legitimate issues that traitsui needs to fix. In places where the warning is not appropriate or if it is getting in the way of releases, suppressing the warnings **in traitsui** is also an option.

I'm getting 19 `UserWarning`s just by importing `from traitsui.api import Color`. Don't think it's a good idea to have this in the releases but don't know where it should be fixed. 
<details>
<summary> Warnings </summary>

```
>>> from traitsui.api import Color
/Users/icernyte/Documents/repos/traitsui/traitsui/handler.py:60: UserWarning: The attribute named 'trait_view_for' of class Handler uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class Handler(HasPrivateTraits):
/Users/icernyte/Documents/repos/traitsui/traitsui/handler.py:531: UserWarning: The attribute named 'trait_context' of class Controller uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class Controller(Handler):
/Users/icernyte/Documents/repos/traitsui/traitsui/handler.py:589: UserWarning: The attribute named 'trait_context' of class ModelView uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ModelView(Controller):
/Users/icernyte/Documents/repos/traitsui/traitsui/ui.py:66: UserWarning: The attribute named 'traits_init' of class UI uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class UI(HasPrivateTraits):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/text_editor.py:64: UserWarning: The attribute named 'traits_view' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/boolean_editor.py:60: UserWarning: The attribute named 'traits_view' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/button_editor.py:36: UserWarning: The attribute named 'traits_view' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/color_editor.py:39: UserWarning: The attribute named 'traits_view' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/file_editor.py:43: UserWarning: The attribute named 'traits_view' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/instance_editor.py:39: UserWarning: The attribute named 'traits_view' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/list_editor.py:73: UserWarning: The attribute named 'trait_handler' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/list_editor.py:73: UserWarning: The attribute named 'traits_view' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/range_editor.py:53: UserWarning: The attribute named 'traits_view' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
/Users/icernyte/Documents/repos/pyface/pyface/action/traitsui_widget_action.py:16: UserWarning: The attribute named 'trait_context' of class TraitsUIWidgetAction uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class TraitsUIWidgetAction(Action):
/Users/icernyte/Documents/repos/traitsui/traitsui/table_column.py:304: UserWarning: The attribute named 'traits_view' of class ObjectColumn uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ObjectColumn(TableColumn):
/Users/icernyte/Documents/repos/traitsui/traitsui/table_column.py:621: UserWarning: The attribute named 'traits_view' of class ListColumn uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ListColumn(TableColumn):
/Users/icernyte/Documents/repos/traitsui/traitsui/table_filter.py:66: UserWarning: The attribute named 'traits_view' of class TableFilter uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class TableFilter(HasPrivateTraits):
/Users/icernyte/Documents/repos/traitsui/traitsui/table_filter.py:408: UserWarning: The attribute named '_trait_values' of class RuleTableFilter uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class RuleTableFilter(TableFilter):
/Users/icernyte/Documents/repos/traitsui/traitsui/editors/table_editor.py:62: UserWarning: The attribute named 'traits_view' of class ToolkitEditorFactory uses one of the reserved prefixes: trait,_trait. Consider renaming it.
  class ToolkitEditorFactory(EditorFactory):
```
</details>

kitchoi:
Thank you @ievacerny for investigating this. It is good to see these warnings.

I am interested in these warnings because I have an interest for traitsui. The following might be best be moved to the traitsui repo...

traitsui has the following options:
(1) Carry on supporting keeping some of these attributes as the public interface, and hope that won't cause interaction with traits like the one seen in #983. Then add a `warnings.filterwarnings` to suppress warnings for the known attribute names, before the offending classes are defined.
(2) Decide that some of these attributes can be deprecated, and recommend a way for users to move to alternative.
(3) Decide that the attributes can be renamed without affecting the user, e.g. the "_trait_values" on `RuleTableFilter` looks like one that can simply be renamed.

PR #1019, opened 2020-04-21
DOC: Clarify when default initializer is called and when handlers are registered

kitchoi:
The behaviour seen in #94 wasn't obvious to me. This PR expands the documentation to clarify:
(1) instantiation with a specific value will be considered for firing a trait change event, if there is a change handler
(2) `post_init` can be used for use cases when firing change handler for values provided to `__init__` is not desirable. This `post_init` was never mentioned in the user documentation, but often mentioned in issues when this comes up.

**Checklist**
- ~Tests~

### DIFF
--- a/docs/source/traits_user_manual/advanced.rst
+++ b/docs/source/traits_user_manual/advanced.rst
@@ -43,9 +43,8 @@ The method overrides any default value specified in the trait definition. The
 initializer is called when:
 
 1. the attribute value is accessed the first time or
-2. an instance is constructed with a value different from the default, if there
-   is a change handler defined for the trait
-   (see :ref:`static-notification`).
+2. an instance is constructed with a specific value, if there is a change
+   handler defined for the trait (see :ref:`static-notification`).
 
 .. index:: get_default_value()
 
--- a/docs/source/traits_user_manual/advanced.rst
+++ b/docs/source/traits_user_manual/advanced.rst
@@ -31,16 +31,21 @@ Dynamic Initialization
 When you define trait attributes using predefined traits, the Trait() factory
 function or trait handlers, you typically specify their default values
 statically. You can also define a method that dynamically initializes a trait
-attribute the first time that the attribute value is accessed. To do this, you
-define a method on the same class as the trait attribute, with a name based on
-the name of the trait attribute:
+attribute. To do this, you define a method on the same class as the trait
+attribute, with a name based on the name of the trait attribute:
 
 .. index:: default value; method
 
 .. method:: _name_default()
 
 This method initializes the *name* trait attribute, returning its initial value.
-The method overrides any default value specified in the trait definition.
+The method overrides any default value specified in the trait definition. The
+initializer is called when:
+
+1. the attribute value is accessed the first time or
+2. an instance is constructed with a different value provided to the instance
+   constructor, if there is a change handler defined for the trait
+   (see :ref:`trait-notification`).
 
 .. index:: get_default_value()
 
--- a/docs/source/traits_user_manual/advanced.rst
+++ b/docs/source/traits_user_manual/advanced.rst
@@ -45,7 +45,7 @@ initializer is called when:
 1. the attribute value is accessed the first time or
 2. an instance is constructed with a different value provided to the instance
    constructor, if there is a change handler defined for the trait
-   (see :ref:`trait-notification`).
+   (see :ref:`static-notification`).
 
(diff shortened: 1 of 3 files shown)


docs/source/traits_user_manual/notification.rst
```diff
@@ -453,21 +453,24 @@ not belong to the same object.
...
+By default, decorated handlers are registered prior to setting the object
+states. When an instance is constructed with a trait value that is different
+from the default, that is considered a change and will fire the associated
+change handlers. The ``post_init`` argument in @on_trait_change can be used
+to delay registering the handler to after the states are set.
 
-A consequence of this difference is that the @on_trait_change() decorator
-causes any default initializers for the traits it references to be executed at
-instance construction time. In the case of specially-named handlers, any
-default initializers are executed lazily.
```
kitchoi, line 470: The last sentence is removed because the behaviour is actually consistent between specially named handlers and decorated handlers: The default initializers are called.

kitchoi: > The default initializers are called.

If the `__init__` is given a value too. But not if `__init__` is not given a value.

docs/source/traits_user_manual/advanced.rst
```diff
@@ -31,16 +31,20 @@ Dynamic Initialization
...
 .. method:: _name_default()
 
 This method initializes the *name* trait attribute, returning its initial value.
-The method overrides any default value specified in the trait definition.
+The method overrides any default value specified in the trait definition. The
+initializer is called when:
+
+1. the attribute value is accessed the first time or
+2. an instance is constructed with a specific value, if there is a change
+   handler defined for the trait (see :ref:`static-notification`).
```
corranwebster, line 47: You might add something about the "why" here: it's needed to provide the "old" value to the trait notification system.

docs/source/traits_user_manual/notification.rst
```diff
@@ -453,21 +453,24 @@ not belong to the same object.
 
 .. _decorator-semantics:
 
-Decorator Semantics
-:::::::::::::::::::
-
+Registration time
```
corranwebster, line 456: Not sure the title needs to change here, but if you do, you probably should update the id of the section on line 454, and also fix any references to the id elsewhere in the document.

docs/source/traits_user_manual/notification.rst
```diff
@@ -453,21 +453,24 @@ not belong to the same object.
...
 The functionality provided by the @on_trait_change() decorator is identical to
 that of specially-named handlers, in that both result in a call to
 on_trait_change() to register the method as a notification handler. However,
 the two approaches differ in when the call is made. Specially-named handlers
 are registered at class construction time; decorated handlers are registered at
-instance creation time, prior to setting any object state.
+instance creation time.
+
+By default, decorated handlers are registered prior to setting the object
+states. When an instance is constructed with a trait value that is different
```
corranwebster, line 467: Nit: I think "state" would be clearer:
```suggestion
state. When an instance is constructed with a trait value that is different
```

docs/source/traits_user_manual/notification.rst
```diff
@@ -453,21 +453,24 @@ not belong to the same object.
...
+from the default, that is considered a change and will fire the associated
+change handlers. The ``post_init`` argument in @on_trait_change can be used
+to delay registering the handler to after the states are set.
```
corranwebster, line 470: Same here:
```suggestion
to delay registering the handler to after the state is set.
```

docs/source/traits_user_manual/notification.rst
kitchoi: Agreed the title can be changed back. Will do that. Thanks!

examples/tutorials/doc_examples/examples/post_init_notification.py
```diff
@@ -0,0 +1,28 @@
...
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+# post_init_notification.py --- Example of static notification
```
kitchoi, line 11: Just spotted this. I changed `"dynamic notification"` to `"static notification"`... to be consistent with the doc.

PR #1023, opened 2020-04-23
Implement logic for adding/removing notifiers for a given ObserverGraph

### LINKED ISSUE #1021: Implement logic for walking down the ObserverGraph and attach/detach notifiers
Part of #977

This issue is for implementing a function for adding/removing notifiers on objects for a given `ObserverGraph` and user's change handler.

The function was implemented here: https://github.com/enthought/traits/blob/observer-framework-test/traits/observers/_observe.py
It will need to be cleaned up and properly tested.

kitchoi:
Closes #1021
Item 6 of #977

This PR adds a function `add_or_remove_notifiers` for adding/removing notifiers for a given `ObserverGraph`.

- The `IObserver` interface is expanded to support `add_or_remove_notifiers`.
- Added a `INotifier` interface to support `add_or_remove_notifiers`. The `TraitEventNotifier` (in #1000) and `ObserverChangeNotifier` (in #1007) implement the `INotifier` interface introduced here. It is possible that this PR might need changes per review outcome of #1000 and #1007. The formal registration can happen later... or not happen at all.

**Checklist**
- [x] Tests
- ~Update API reference (`docs/source/traits_api_reference`)~: Things here are still private. Module names are preceded with an underscore.
- ~Update User manual (`docs/source/traits_user_manual`)~: Things here are still private. 
- ~Update type annotation hints in `traits-stubs`~: Nothing here are trait-ed

### DIFF
--- a/traits/observers/_i_notifier.py
+++ b/traits/observers/_i_notifier.py
@@ -1,0 +1,42 @@
+
+import abc
+
+
+class INotifier(abc.ABC):
+    """ Interface for all notifiers.
+
+    An instance of notifier must be a callable, i.e. ``__call__`` must be
+    implemented and cannot be None. The signature of that callable should be
+    compatible with the observables the notifier will be given to. This
+    interface does not define what that signature should be.
+    """
+
+    def __call__(self, *args, **kwargs):
+        """ Called by the observable.
+        The signature is not restricted by the interface.
+        """
+        raise NotImplementedError("__call__ must be implemented.")
+
+    def add_to(self, observable):
+        """ Add this notifier to the observable.
+
+        Parameters
+        ----------
+        observable : IObservable
+        """
+        raise NotImplementedError("add_to must be implemented.")
+
+    def remove_from(self, observable):
+        """ Remove this notifier or a notifier equivalent to this one
+        from the observable.
+
+        Parameters
+        ----------
+        observable : IObservable
+
+        Raises
+        ------
+        NotifierNotFound
+            If the notifier cannot be found.
+        """
+        raise NotImplementedError("remove_from must be implemented.")
--- a/traits/observers/_i_notifier.py
+++ b/traits/observers/_i_notifier.py
@@ -12,7 +12,7 @@ class INotifier(abc.ABC):
     """
 
     def __call__(self, *args, **kwargs):
-        """ Called by the observable.
+        """ Called by an observable.
         The signature is not restricted by the interface.
         """
         raise NotImplementedError("__call__ must be implemented.")
--- a/traits/observers/_i_notifier.py
+++ b/traits/observers/_i_notifier.py
@@ -1,3 +1,12 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
 
 import abc
 
--- a/traits/observers/_i_observer.py
+++ b/traits/observers/_i_observer.py
@@ -28,8 +28,7 @@ class IObserver(abc.ABC):
     to the leaf nodes. The first observer (root node) will be given the object
     the user wants to observe. It knows how to obtain observable(s) for
     attaching notifiers. It also knows what objects should be given to the
-    downstream/children observers in the same graph. Then the process repeats
-    as the ``ObserverGraph`` is walked.
+    downstream/children observers in the same graph.
 
     An observer defines the notifier that wraps the user change handler, should
     notification is enabled by the user. This allows the observer to define
--- a/traits/observers/_i_observer.py
+++ b/traits/observers/_i_observer.py
@@ -80,11 +80,6 @@ class IObserver(abc.ABC):
         Yields
         ------
         IObservable
-
-        Raises
-        ------
-        ValueError
-            If the given object cannot be handled by this observer.
         """
         raise NotImplementedError("iter_observables must be implemented.")
 
--- a/traits/observers/_i_observer.py
+++ b/traits/observers/_i_observer.py
@@ -20,6 +20,21 @@ class IObserver(abc.ABC):
     user. In order to support equality and hashing on the
     ``ObserverGraph``, ``IObserver`` needs to be hashable
     and it needs to support comparison for equality.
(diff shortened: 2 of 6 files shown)


traits/observers/_i_notifier.py
```diff
@@ -0,0 +1,51 @@
...
+        """ Remove this notifier or a notifier equivalent to this one
+        from the observable.
+
+        Parameters
+        ----------
+        observable : IObservable
+
+        Raises
+        ------
+        NotifierNotFound
```
kitchoi, line 48: This `NotifierNotFound` is in #1000.

kitchoi: Introduced the `NotifierNotFound` here too...resolving the merge conflict is trivial, remembering to bring it back to tests is harder (for me anyway).

traits/observers/_i_observer.py
```diff
@@ -29,3 +44,104 @@ def __hash__(self):
...
         """ Return true if this observer is equal to the given one."""
         raise NotImplementedError("__eq__ must be implemented.")
+
+    @property
+    def notify(self):
+        """ A boolean for whether this observer will notify for changes.
+        """
+        raise NotImplementedError("notify property must be implemented.")
+
+    def iter_observables(self, object):
```
ievacerny, line 54: I find the `iter_*` interface really confusing. I had to look at interface implementations to understand what exactly each one is iterating over just to find that `iter_observables` and `iter_extra_graphs` almost always return only one object. And with just having this interface description it's really hard to understand what `add_or_remove_notifiers` is doing as well.

I don't know how to fix this or even if it's worth fixing. Just wanted to bring this up, maybe someone else has any ideas.

kitchoi: Thank you for bringing this up. I am also not sure how to fix this, maybe it just needs a different docstring?

Let me try to explain: For these `iter_*`, the interface is asking the implementation to "give me zero or many observables", therefore these are generators.

For example, if we want to observe traits that have a specific metadata, this `iter_observables` may yield nothing, if no traits have the required metadata, or many observables, if an instance of `HasTraits` has more than one trait that has the required metadata. Similarly for `iter_objects`.

For `iter_extra_graphs`, so far it is only needed for adding an observer for the `trait_added` event.  Some observers don't need the handling `trait_added` event, and their `iter_extra_graphs` will yield nothing. Other observer will only yield one extra graph for handing `trait_added`.
An alternative to this is to define a method that will return an instance of ObserverGraph or None for no graphs. But then `add_or_remove_notifiers` does not need to care whether there must be at most one extra graphs. It can deal with many graphs just fine. So a generator to ask for zero or many things seem to be the most flexible (then `add_or_remove_notifiers` also doesn't need to check for `None`).

kitchoi: > "give me zero or many observables things"

To be precise, I mean a minimum of zero things, with no maximum bound. The count of one is allowed here too.

kitchoi: Maybe some examples would help:

`iter_obserbables` yielding one thing for observing a specifically named trait:
https://github.com/enthought/traits/blob/1ea851ad118d054096ee13cd1e57a6d592fe81a2/traits/observers/_named_trait_observer.py#L80-L103

`iter_observables` yielding potentially nothing, but maybe many things when observing traits filtered by some callable (e.g. existence of a metadata):
https://github.com/enthought/traits/blob/1ea851ad118d054096ee13cd1e57a6d592fe81a2/traits/observers/_filtered_trait_observer.py#L64-L91

`iter_extra_graphs` yielding nothing for observing items on a list, because there is no "add_trait" action on a TraitList(TraitListObject to be precise):
https://github.com/enthought/traits/blob/1ea851ad118d054096ee13cd1e57a6d592fe81a2/traits/observers/_list_item_observer.py#L202-L215

`iter_extra_graphs` yielding one thing for handling `HasTraits.add_trait` as the observer is given an instance of `HasTraits`:
https://github.com/enthought/traits/blob/1ea851ad118d054096ee13cd1e57a6d592fe81a2/traits/observers/_named_trait_observer.py#L171-L190

ievacerny: Examples would definitely help - looking at implementation is how I eventually understood what each generator is for. I'm just wondering if they might become a maintenance problem in the future 🤔

kitchoi: I have expanded the docstring to mention some examples. See what you think. I hope to keep it generic enough so that examples don't get confused as rules.

I think it would also help when the concrete implementations are added to the code base, I can't do that yet because they are going to depend on the notifier that is still being reviewed.

traits/observers/_i_observer.py
```diff
@@ -20,6 +20,21 @@ class IObserver(abc.ABC):
...
+    to the leaf nodes. The first observer (root node) will be given the object
+    the user wants to observe. It knows how to obtain the ``IObservable``
+    object(s) for attaching notifiers (see ``iter_observables``). It also
+    knows what objects should be given to the downstream/children observers in
+    the same graph (see ``iter_objects``). Then the process repeats as the
+    ``ObserverGraph`` is walked. See ``add_or_remove_notifiers``.
+
+    Some observers may require additional helper observers and notifiers,
+    e.g. an observer for traits will need to respond to the ``trait_added``
+    event. See ``iter_extra_graphs``.
```
kitchoi, line 37: Now this sentence may be a bit redundant given the recent change to docstring of `iter_extra_graphs`. Will remove.

traits/observers/_i_observer.py
ievacerny: New docstrings look good! Thanks for that.

As per out discussion on slack, I was trying to think of a better phrasing for the first sentence of `iter_objects` docstring but came up with nothing. I think warnings on implementations that deal with CTrait should be enough.

kitchoi: SGTM!  Thank you for hashing this out!

traits/observers/_i_observer.py
```diff
@@ -29,3 +43,142 @@ def __hash__(self):
...
+            Concrete should do the sanity check where appropriate.
+
+        Yields
+        ------
+        IObservable
+
+        Raises
+        ------
+        ValueError
+            If the given object cannot be handled by this observer.
```
kitchoi, line 86: This Raises section is not part of the interface. Implementation is allowed to raise whatever type of exceptions they like.  See the `try-except-Exception-re-raise` in `_AddOrRemoveNotifier.__call__`. I will remove this section from the docstring.

kitchoi:
Thank you @ievacerny and @corranwebster!
Merging...

PR #1069, opened 2020-05-07
Implement the logic for observing a named trait

kitchoi:
This PR implements item 7 in #977 for observing a named trait.

- ~Add `TraitChangeEvent` for wrapping `(object, name, old, new)`. This is the object to be received by the user handler.~ Edited: See #1070
- Add `optional` argument in order to support `"items"` special keyword in the mini-language. The `"items"` keyword is ambiguous about whether it means a trait named "items" or items in a container. We will use this named trait observer but set optional to true for that usage.
- Extend `NamedTraitObserver` to implement `IObserver` interface:
      - return the instance Ctrait for adding/removing notifiers (including "maintainer")
      - return the value for the next observer, if it is defined. Care is taken to avoid evaluating default values and passing on values such as `Undefined`.
      - return a notifier wrapping the user's change handler so that the handler receives `TraitChangeEvent`
      - return a notifier (the "maintainer") for maintaining downstream observers when the trait change.

When an observer is registered, it should not trigger default initializers to be evaluated. That is consistent with the current behaviour of `on_trait_change`. However the implementation in `on_trait_change` for avoiding evaluating defaults caused other issues (e.g. #275) which aren't easy to fix. The implementation here will not cause the same issue as seen in #275 and will avoid the side effect of evaluating default at the same time.

Not in this PR:
- `trait_added` event is not included here. That is in item 11 of #977. (I also don't want this PR to be any bigger).

Note that in #977 there will also be a "FilteredTraitObserver" which can be used for observing traits with an arbitrary filter function.  While `NamedTraitObserver(..., optional=True)` could have been implemented using `FilteredTraitObserver`, `NamedTraitObserver(..., optional=False)` cannot. Raising an error when a requested trait is not found will allow us to avoid issues like #447, it will also be useful for catching human errors or for migrating from `on_trait_change` to `observer`, as the mini-language syntax has changed.

**Checklist**
- [x] Tests
- [x] Update API reference (`docs/source/traits_api_reference`): Will conflict with #1067 but the conflicts are trivial.


traits/observers/_has_traits_helpers.py
```diff
@@ -0,0 +1,114 @@
...
+
+    if event.new not in UNOBSERVABLE_VALUES:
+        add_or_remove_notifiers(
+            object=event.new,
+            graph=graph,
+            handler=handler,
+            target=target,
+            dispatcher=dispatcher,
+            remove=False,
+        )
```
kitchoi, line 114: Functions in these modules are refactored as they will be reused by `FilteredTraitObserver`.

traits/observers/_named_trait_observer.py
```diff
@@ -43,12 +60,153 @@ def notify(self):
...
+        Returns
+        -------
+        notifier : TraitEventNotifier
+        """
+        return TraitEventNotifier(
+            handler=handler,
+            target=target,
+            dispatcher=dispatcher,
+            event_factory=trait_event_factory,
+            prevent_event=lambda event: event.old is Uninitialized,
```
kitchoi, line 161: This is consistent with the current behaviour of `on_trait_change`.

kitchoi:
I moved the `TraitChangeEvent` to a separate PR in #1070 because that part is easier to review. Having that event object in the code base will allow me to get a few other PRs going in parallel.

traits/observers/tests/test_named_trait_observer.py
```diff
@@ -65,20 +97,282 @@ def test_observer_graph_hash_with_named_listener(self):
         # Test equality + hashing using set passes.
 
         path1 = ObserverGraph(
-            node=NamedTraitObserver(name="foo", notify=True),
+            node=create_observer(name="foo"),
             children=[
-                ObserverGraph(
-                    node=NamedTraitObserver(name="bar", notify=True),
-                ),
+                create_observer(name="bar"),
```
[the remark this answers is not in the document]
kitchoi: Oops, yes you are right, thank you!

traits/observers/_named_trait_observer.py
```diff
@@ -43,12 +60,153 @@ def notify(self):
         """
         return self._notify
 
+    @property
+    def optional(self):
```
corranwebster, line 64: Thought: should this class be a subclass of named tuple`?  This would get you a lot of the infrastructure for free (you wouldn't need properties and shadow attributes, `__hash__` will just work, and `__eq__` may also just work or need minimal wrapping.

kitchoi:
Thank you both for your review!

On namedtuple: I would want to use it if it does not also throw in `__iter__`, __contains__`, `__len__` and friends. Those are features not needed by these observer classes. By not depending on more than needed, this allows more freedom for future changes. I don't mind the boilerplate so much, it allows ample space for documentation.

PR #1060, opened 2020-05-01
Add user manual for observe

kitchoi:
This PR add user manual for the `observe` framework.

- The documentation for `on_trait_change` is moved to `listening.rst`. The first paragraph is extracted as an intermediate index page for notifications.
- Add `observing.rst` for the `observe` framework.
- Add `traits.observation.api` to the API reference for cross-referencing.
- Add examples to `examples/tutorials/doc_examples`

Edited (this was the previous message:)
~This is a draft PR for making the `observe` user manual visible (see https://github.com/enthought/traits/issues/977).~

~Much of the API cross-references are missing in this PR, as the PR would otherwise be too big. To see the documentation with styling and cross-referencing, checkout `observer-framework-test` branch and build the documentation using `python etstool.py docs`.~

**Checklist**
- ~Tests~
- [x] Update API reference (`docs/source/traits_api_reference`)
- [x] Update User manual (`docs/source/traits_user_manual`)
- ~Update type annotation hints in `traits-stubs`~

PR #1070, opened 2020-05-08
Add TraitChangeEvent object for wrapping CTrait change

kitchoi:
This PR adds the `TraitChangeEvent` object for wrapping CTrait change events.

This event object will be given to the user's handler registered using `observe`.

This event object will also be used by the `NamedTraitObserver` in #1069, the "FilteredTraitObserver" for observing traits using a filter, and the (internal-use-only) observer for `trait_added` events.

**Checklist**
- [x] Tests
- ~Update User manual (`docs/source/traits_user_manual`)~ WIP in #1060 

### DIFF
--- a/docs/source/traits_api_reference/index.rst
+++ b/docs/source/traits_api_reference/index.rst
@@ -40,6 +40,7 @@ Subpackages
 
     traits.adaptation
     traits.etsconfig
+    traits.observers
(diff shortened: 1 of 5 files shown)


kitchoi:
Thank you! Merging this and I can merge master back to #1069

PR #1071, opened 2020-05-08
Implement ListItemObserver for observing mutations to a list

kitchoi:
This PR implements item 8 of #977:

> Implement the logic for attaching notifiers when the items of a list changes

- Add `ListChangeEvent` for representing mutations to a list. This object will be given to the user's change handler.
- `ListItemObserver` implements the interface of `IObserver` to:
     - return the object for adding/removing notifiers. In this case, the observable is simply an instance of `TraitList`. If the incoming object is not a TraitList and the observer is not marked as optional, the observer will complain. This is to help catch usage errors.
     - yield the content of the list for the next observer(s) in an ObserverGraph. The `ListItemObserver` has no knowledge about the nature of the next observers (this is generally true for any observers). If the content of the list is not compatible with the next observer (say the list contains integers), the next observer is expected to complain about it. e.g. see `NamedTraitObserver` in #1069 which raises an error when the incoming object is not an instance of HasTraits.
- Implement `IObservable` interface in `TraitList`.  With this, any subclass of `TraitList`, e.g. TraitListObject, can be observed with `ListItemObserver` (see tests).

**Checklist**
- [x] Tests


traits/observers/tests/test_list_item_observer.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def test_not_equal_optional(self):
+        observer1 = ListItemObserver(notify=True, optional=True)
+        observer2 = ListItemObserver(notify=True, optional=False)
+        self.assertNotEqual(observer1, observer2)
+
+    def test_not_equal_different_type(self):
+        observer1 = ListItemObserver(notify=False, optional=False)
+        imposter = mock.Mock()
+        imposter.notify = False
```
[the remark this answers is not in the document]
mdickinson: Nit: `impostor`, not `imposter`. (Sorry.)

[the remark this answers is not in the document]
mdickinson: Sigh. Nope; both are correct.

traits/observers/_list_item_observer.py
```diff
@@ -0,0 +1,219 @@
...
+        parser, where the source container type is ambiguous.
+    """
+
+    def __init__(self, *, notify, optional):
+        self._notify = notify
+        self._optional = optional
+
+    def __hash__(self):
+        """ Return a hash of this object."""
+        return hash((type(self), self.notify))
```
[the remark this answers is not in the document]
kitchoi: Well spotted. That would be a nasty bug. Thank you!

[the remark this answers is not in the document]
kitchoi: Actually, it would not be that too bad because equal hash does not mean equal objects. It is for this reason I could not write a failing test for this.

traits/observers/_list_change_event.py
```diff
@@ -0,0 +1,72 @@
...
+# ListChangeEvent is exposed in the public API
+
+
+class ListChangeEvent:
+    """ Event object to represent mutations to a list.
+
+    Attributes
+    ----------
+    trait_list : list
+        The list being mutated.
```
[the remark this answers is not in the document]
kitchoi: Whether the list contains content before or after the change depends on whether the change event is fired before or after the mutation, because it is the same list object. Strictly speaking, the event object does not need to dictate when the notification should be fired. Not sure how I could write this...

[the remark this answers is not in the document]
kitchoi: We could add to the API document (not here) that notifications are fired after the mutations. What do you think?
i.e. `TraitList.notify` are called after the mutation has taken place.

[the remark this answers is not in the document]
ievacerny: Oh, okay, I thought the behaviour was consistent. I think it would be good to report which events are fired before changes and which are fired after, but I agree that the event itself is not a place for that.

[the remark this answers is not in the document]
kitchoi: I am sorry I did not mean to imply there were events fired before changes. All events are fired after changes.  Sorry for the confusion there.

kitchoi:
Thank you @ievacerny for the review!
I hope I have addressed all the comments.
There will be some trivial merge conflicts with #1069 due to the shared API module and documentation source. I will resolve those when that happens.

traits/observers/_list_change_event.py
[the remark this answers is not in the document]
ievacerny: Then maybe having past tense tense in all 4 attribute descriptions would be a good way to say that the changes have happened without "dictating when notification is fired"?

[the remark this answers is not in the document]
kitchoi: I was meaning to say that the event object is agnostic about whether the change has happened or is about to happen. The "added" and "removed" descriptions are intentionally ambiguous.  We should be able to reuse this object to represent a change that is about to happen (that is not a use case, but there is no need for this event to prevent that).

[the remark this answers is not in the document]
ievacerny: Okay, well I would still argue that it would be good to inform the user that the event is fired after changes are made, but then I really don't know what would be the right place to do so.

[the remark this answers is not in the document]
kitchoi: Yes I agree that the information needs to be somewhere in the documentation. In fact, I would very much like that to be included in the user manual.

While it may be convenient to have that description in the event object parameter section, doing so may introduce restrictions that do not belong to the data structure.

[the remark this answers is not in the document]
kitchoi: Actually, the first sentence in the current user manual already implies that all change notifications are fired after the change:

https://docs.enthought.com/traits/traits_user_manual/notification.html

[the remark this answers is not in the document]
ievacerny: I see that similar implicit mentions are made in the new observer's user manual as well, so it might be enough. I was just a bit worried that because the handlers are now supposed to deal with the actual events, the user's might be paying more attention to the docstrings of those events.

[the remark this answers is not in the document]
corranwebster: It might not be amplified enough, but it is a fairly strong feature of Traits that notifications _always_ fire after the value has set.

traits/observers/_list_change_event.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def __init__(self, *, trait_list, index, removed, added):
+        self.trait_list = trait_list
+        self.added = added
+        self.removed = removed
+        self.index = index
+
+    def __repr__(self):
+        return (
+            "<ListChangeEvent("
```
corranwebster, line 40: I usually format in `self.__class__.__name__` instead of hard-coding the name, so the `repr` continues to be accurate if someone subclasses the class.

traits/observers/_list_change_event.py
```diff
@@ -0,0 +1,72 @@
...
+            ")>".format(
+                trait_list=self.trait_list,
+                index=self.index,
+                removed=self.removed,
+                added=self.added,
+            )
+        )
+
+
+def list_event_factory(trait_list, index, removed, added):
```
corranwebster, line 52: I'd be tempted to make this a `classmethod` of `ListChangeEvent`

traits/observers/_list_item_observer.py
```diff
@@ -0,0 +1,221 @@
...
+from traits.observers._list_change_event import list_event_factory
+from traits.observers._i_observer import IObserver
+from traits.observers._trait_event_notifier import TraitEventNotifier
+from traits.observers._observe import add_or_remove_notifiers
+from traits.observers._observer_change_notifier import ObserverChangeNotifier
+from traits.trait_list_object import TraitList
+
+
+@IObserver.register
+class ListItemObserver:
```
corranwebster, line 20: As in the last PR, can this be a subclass of named tuple?

Also had the thought that even if you don't turn it into a named tuple, using `__slots__` here might make sense for efficiency (keep that in the back of your mind, even if it isn't part of this PR).

traits/observers/_list_change_event.py
```diff
@@ -0,0 +1,72 @@
...
+#
+# Thanks for using Enthought open source!
+
+""" Event object for representing mutations to a list.
+"""
+
+# ListChangeEvent is exposed in the public API
+
+
+class ListChangeEvent:
```
corranwebster, line 17: Possibly make this a named tuple?

traits/observers/_list_change_event.py
corranwebster: Actually, on second thought, perhaps embrace the ordering and make `ListChangeEvent` a named tuple.

traits/observers/_list_change_event.py
```diff
@@ -0,0 +1,72 @@
...
+        The list being mutated.
+    index : int or slice
+        The index used for the mutation.
+    added : list
+        Values added to the list.
+    removed : list
+        Values removed from the list.
+    """
+
+    def __init__(self, *, trait_list, index, removed, added):
```
corranwebster, line 32: Is there are particular reason you want these to be keyword only?

traits/observers/_list_change_event.py
```diff
@@ -0,0 +1,72 @@
...
+
+# ListChangeEvent is exposed in the public API
+
+
+class ListChangeEvent:
+    """ Event object to represent mutations to a list.
+
+    Attributes
+    ----------
+    trait_list : list
```
corranwebster, line 22: Also, a design suggestion for this and the other collection change events: it might make sense to try to mimic the API of `TraitChangeEvent`.  At a minimum:

- using `object` instead of `trait_list`
- a dummy attribute `name` which is always `"items"`

This might allow code to consume both types of event without having to do a type check if it only cares about the object and the name.

That would imply an implicit "ChangeEvent" interface where you expect both "object" and "name" to exist.  It is tempting to make `new` be the new value of the list (ie. the same as `object`) and `old` to be a property that lazily reconstructs the old value of the list from the `added` and `removed`, but that isn't really right from a philosophical point of view.

Really the rest of the information is the "delta", ie. we have
- "who changed": the object that was changed, be it a HasTraits, TraitList, etc.
- "what changed": the name of the thing that changed (the thing that was being observed)
- "how it changed": the nature of the changes that took place (old/new, added/removed, etc.)

Thinking this way may help with the design of these classes.

kitchoi: Hmmm we might need some use cases to drive this. For example, if a handler needs to consume `object` and `name`, the next thing it does may be to call `getattr(object, name)`, and doing that with `object` being `TraitList` and `name` being `"items"` is going to fail.

If the handler needs to do something specific about the change details, then it probably needs to know which type of event objects it gets anyway, it could go two routes:
(1) Apply `notify=True` only on the type of change that is compatible. Then it only deals with one type of event objects.
(2) Check the event type, and do different things according to the type.

Are there use cases already we can draw conclusions from?  Otherwise, leaving these `object` and `name` undefined for now might actually be more flexible, so we don't have to commit to a value before knowing how they are going to be used. e.g. `object` could be the `HasTraits` instance that contains the list somewhere (note that the list could be nested). It does not hurt if we end up having both `object` and `trait_list` pointing to the same thing.

traits/observers/_list_change_event.py
kitchoi: Enforcing keyword-only arguments will help prevent bugs like this one: https://github.com/enthought/traitsui/issues/791

traits/observers/_list_change_event.py
kitchoi: The `__init__` of `ListChangeEvent` is currently the same as the call signature of `TraitList.notify` now, but they could evolve independently of each other.  `DictChangeEvent` in #1072 is actually an example where the `__init__` of `DictChangeEvent` is not the same as the call signature of `TraitDict.notify`.

The only reason why this is not a classmethod is to avoid exposing this to the user. `ListChangeEvent` is exposed in the public API via the public module `events` whereas the event factory isn't. An alternative organization would be to have all the public facing event object defined in a public module, and these event factories defined in another private modules.

traits/observers/_list_change_event.py
kitchoi: Good point. Will do that.

traits/observers/_list_change_event.py
kitchoi: > The __init__ of ListChangeEvent is currently the same as the call signature of TraitList.notify

That's not true because of the keyword arguments...and of course that's why you asked in the next comment :sweat_smile:

traits/observers/_list_item_observer.py
kitchoi: I did give this a go, and while it got rid of the `__init__`, `__hash__`, `__eq__` and the properties, the equality check does not quite do what is implemented here. e.g. the following returns true.
```
ListItemObserver(notify=True, optional=False) == (True, False)
```
That is not very convenient considering that `ListItemObserver`, `SetItemObserver` and `DictItemObserver` have the same attributes and they should not be equal to each other.

traits/observers/_list_change_event.py
kitchoi: namedtuple turns this into a sequence-like object. Unlike the observer, this is a public-facing object, so it would be difficult going back. Maybe better to wait until we really need to.

mdickinson: I agree about avoiding namedtuple; it's an interface that presents multiple ways to do it, which often isn't what you want. It's useful as a transitional type when going from something that was previously a tuple to something closer to a class with attributes, but I don't think it's what we want here. A dataclass may be closer to what we want, but that's not available until Python 3.7.

kitchoi:
I have addressed the ones that I could address. There are some pending discussion points and I am not sure what to do with those.

traits/observers/_list_item_observer.py
mdickinson: Let's keep `__slots__` away from this for now; using `__slots__` can have some interesting side-effects (particularly with regard to subclassing). It's an optimization that we should save until we need it (and my bet would be that we won't ever need it).

traits/observers/_list_change_event.py
mdickinson: I agree that keyword-only arguments make sense here: it makes the calling code clearer as well, particularly with respect to the ordering of `removed` and `added` - someone reading the client code doesn't have to check or remember which way around the arguments go.

The keyword-only restriction is also something that's easy to relax if it becomes problematic. It's harder to add the restriction back in if it's not present to begin with.

traits/observers/_list_change_event.py
```diff
@@ -0,0 +1,73 @@
...
```
mdickinson, line 22: Can we be more precise with the type here? Will `trait_list` always be an instance of `TraitList`?

traits/observers/_list_change_event.py
```diff
@@ -0,0 +1,73 @@
...
+        self.removed = removed
+        self.index = index
+
+    def __repr__(self):
+        return (
+            "<{class_name}("
+            "trait_list={trait_list!r}, index={index!r}, "
+            "removed={removed!r}, added={added!r}"
+            ")>".format(
+                class_name=type(self).__name__,
```
mdickinson, line 44: Suggestion: the `format` operation could be simplified by passing `self=self`, and then using `self.trait_list`, `self.index` and so on in the template.

traits/observers/_list_item_observer.py
```diff
@@ -0,0 +1,221 @@
...
+        parser, where the source container type is ambiguous.
+    """
+
+    def __init__(self, *, notify, optional):
+        self._notify = notify
+        self._optional = optional
+
+    def __hash__(self):
+        """ Return a hash of this object."""
+        return hash((type(self), self.notify, self.optional))
```
mdickinson, line 40: The use of `type(self)` here makes me a bit uncomfortable: it introduces an extra level of non-deterministic behaviour, since memory address, and hence the hash, of the type will likely vary from run to run. How about using the type's name instead? That at least gives us the option of making things determinstic (e.g., during debugging) by setting `PYTHONHASHSEED`.

traits/observers/_list_item_observer.py
```diff
@@ -0,0 +1,221 @@
...
+    def __eq__(self, other):
+        """ Return true if this observer is equal to the given one."""
+        return (
+            type(self) is type(other)
+            and self.notify == other.notify
+            and self.optional == other.optional
+        )
+
+    @property
+    def notify(self):
```
mdickinson, line 51: What's the rationale for putting these attributes behind properties? I'd suggest keeping it simple and leaving them as plain attributes.

traits/observers/_list_item_observer.py
```diff
@@ -0,0 +1,221 @@
...
+        Description for the *downstream* observers, i.e. excluding self.
+    handler : callable
+        User handler.
+    target : object
+        Object seen by the user as the owner of the observer.
+    dispatcher : callable
+        Callable for dispatching the handler.
+    """
+    for removed_item in event.removed:
+        add_or_remove_notifiers(
```
mdickinson, line 205: The `add_or_remove_notifiers(..., remove=True)` feels awkward here; I think we probably want `add_notifiers` and `remove_notifiers` functions instead, for clarity and simplicity. (This is one of Guido's rules for Pythonic interfaces: if you have a boolean flag that's almost always passed as a constant in practice, you probably want two separate functions instead.)

traits/observers/tests/test_list_change_event.py
```diff
@@ -0,0 +1,62 @@
...
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+import unittest
+
+from traits.trait_list_object import TraitList
+from traits.observers._list_change_event import (
```
mdickinson, line 14: nitpick: :abc:  Yes, I know we're not consistent about import order in existing code, but in new code please keep things in alphabetical order, so that we can slowly move _towards_ consistency.

traits/observers/_list_change_event.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def __init__(self, *, trait_list, index, removed, added):
+        self.trait_list = trait_list
+        self.added = added
+        self.removed = removed
+        self.index = index
+
+    def __repr__(self):
+        return (
+            "<{class_name}("
```
mdickinson, line 40: Let's get rid of the `<` and `>` enclosing brackets here. Those are traditionally only used for `<... object at 0xdeadbeef>`-style representations. Here we're aiming for something that's `eval`-able back to the existing object.

traits/observers/tests/test_list_item_observer.py
```diff
@@ -0,0 +1,329 @@
...
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+import unittest
+from unittest import mock
+
+from traits.api import HasTraits, List, Instance, Int
```
mdickinson, line 14: nitpick :abc:

traits/observers/_list_item_observer.py
kitchoi: Initially I prefer not using the name because it assumes two classes must have unique name, but it is not true, e.g. There are two `TraitList` classes, and two `Expression` classes in the traits distribution.

We could include the fully qualified module names though. I think that's what I am going to do.

traits/observers/_list_item_observer.py
kitchoi: I was hoping make these attributes immutable, to be defensive (against human errors/oversights).

traits/observers/_list_change_event.py
mdickinson: > "how it changed": the nature of the changes that took place (old/new, added/removed, etc.)

This is one of the reasons that I think it would make sense to bundle `index`, `removed` and `added` up into a single object, as I've suggested before. Those three pieces of information are always generated together, passed around together and used together.

Maybe open an issue to revisit this once the various related PRs are merged and we have a chance to look at the design as a whole?

traits/observers/_list_item_observer.py
mdickinson: I think that only makes sense in Python where there's a higher-than-usual risk of attributes being overwritten (e.g., perhaps because there was a previous API that did this). For this particular situation, I don't think this type of accident is any more likely than any of the other possible accidents that we're not protecting against, so I think this is unnecessary extra boilerplate.

mdickinson: > a higher-than-usual risk of attributes being overwritten

Or perhaps also in cases where the consequences of accidental overwriting would be worse than usual.

traits/observers/_list_item_observer.py
kitchoi: On the one hand, I agree with that. I prefer `observe` and `unobserve`.

However, that would go against local consistency with `on_trait_change`. This pattern is not just in traits, but also in traitsui and pyface.

On the other hand, this `add_or_remove_notifiers` is only used internally so it would also be fine to separate them.  On the other other hand, it is much easier to ensure the "add" operation and "remove" operation are the exact opposite if they are implemented in the same function. 

All in all, we might end up having an intermediate `add_notifiers` and `remove_notifiers` that wraps an `add_or_remove_notifiers`, and then the `add_notifiers` and `remove_notifiers` being indirectly called by `observe(remove: boolean)`. 🤯

kitchoi: In fact, separate `add_notifiers` and `remove_notifiers` are what I initially had, but then I ended up having another if-else-branch in `observe` wrapping something that has another if-else-branch (the code that does the mutation on the notifier list), so I gave up on that. The pattern has to happen consistently for this to be helpful.

traits/observers/_list_item_observer.py
kitchoi: These attributes are special because they are used in `__hash__`, therefore I am making an exclusive protection against mutating them. I am happy to relax it, just wanted to clarify in case that may change your view, as I realized I have not included the mention of hashing in my previous comment.

traits/observers/_list_change_event.py
kitchoi: I am open to that discussion and it makes sense to do that at the very end instead of blocking these PRs. 
Do you mind opening the said issue please (since I am worried I may misunderstand the scope you are thinking about)?

traits/observers/_list_item_observer.py
mdickinson: > it assumes two classes must have unique name

Why? There's no rule that says that hashes have to be unique - and in fact, they can't be. For practical purposes, all you want is to ensure that for unequal things that are likely to end up in a set or dict together, it's not too likely for those things to have the same hash value. And if they do, everything will still work just fine; it might just be a tiny amount slower thanks to some extra collisions internally.

So it's fine to just use the type name here; there's no need for fully qualified names. (Nothing _wrong_ with using fully qualified names if that's what you want to do, but it doesn't bring any benefit.)

traits/observers/_list_item_observer.py
kitchoi: >  am happy to relax it, just wanted to clarify in case that may change your view, as I realized I have not included the mention of hashing in my previous comment.

Just to clarify that I am not proceeding with this change yet until I get a confirmation.

traits/observers/_list_item_observer.py
kitchoi: That's true.

traits/observers/_list_item_observer.py
mdickinson: Let's simplify the code. I don't think this is a failure mode that's likely enough to make it worth guarding against.

kitchoi: OK. I don't mind the boilerplate and I think defensiveness is good, but I will remove it nonetheless.

mdickinson: Yes, defensiveness is good, but this is one isolated piece of defensiveness that's at a level that's more low risk than many other pieces we're not addressing: it's like having a lock on one window in a house while you leave all the other windows wide open. If we applied this level of defensiveness everywhere there would be extra boilerplate everywhere, and much of the advantage we get from programming in a low boilerplate language would be gone.

traits/observers/_list_change_event.py
mdickinson: > Do you mind opening the said issue please

Opened #1084.

traits/observers/_list_item_observer.py
mdickinson: Okay, let's leave as-is for now.

mdickinson:
@corranwebster gave his okay on another channel. Merging.

PR #1085, opened 2020-05-15
Register all observables to IObservable

kitchoi:
This PR makes sure all observables are (virtual) subclasses of `IObservable`.

So far nothing does the subclassing check e.g. `issubclass` or `isinstance`, so the registeration is arguably unused.
This is more about explicitly declaring a type as observable and making observables more discoverable (e.g. one can do a global search on `IObservable`. Unfortunately registering virtual subclass does not allow reverse lookup from the `ABC`...)

Drive-by import and naming changes to make things more consistent.

**Checklist**
- ~Tests~

### DIFF
--- a/traits/ctrait.py
+++ b/traits/ctrait.py
@@ -20,6 +20,7 @@ import inspect
 
 from . import ctraits
 from .constants import ComparisonMode, DefaultValue, default_value_map
+from .observers._i_observable import IObservable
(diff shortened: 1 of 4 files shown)


kitchoi:
Note that `HasTraits` is excluded here still.  `HasTraits._notifiers` is not used by `observe` anywhere.

PR #1242, opened 2020-07-14
[Draft] Recursion support for observe

kitchoi:
This is a draft PR that will be closed immediately. This is created solely for future reference and to keep a record for work that was done before.

Background: In the two proof-of-concept PRs for the observe framework (#942 and #969), recursion support was experimented with because `on_trait_change` does have this support. The feature was excluded for it is not clear if there is indeed a need, and the implementation isn't trivial.

The main difficulties are:
- We need to compare two graphs containing cycles to determine if they are equivalent (graph isomorphism problem). But given our use case, we can make a few assumptions and shortcuts to make this easier.
- We need to construct an `ObserverGraph` containing cycles from an expression (e.g. in the mini-language `"a.[b,c]*"`). This requires maintaining the identity of objects as the graph is constructed.

This implementation does the following:
- Change the data structure of `ObserverGraph` so it keeps track of what are branches and what are cycles.
- In `Expression`, keeps track of the `ObserverGraph` objects in a cache, it is similar to how `deepcopy` handles cycle references.

Important note:
- Previous work in #969 tried to construct the graph "from inside-out", because "outside-in" did not work then. Here the construction is made from the outside-in. I realized that previously the "outside-in" approach did not work because the `ObserverGraph` was hashing its children graphs too early and the hash was persisted.  It is important that during the process of constructing the observer graph with cycles, the hashes of the intermediate observer graphs are never used or persisted.

**Checklist**
- [x] Tests

### DIFF
--- a/traits/observation/_filtered_trait_observer.py
+++ b/traits/observation/_filtered_trait_observer.py
@@ -170,5 +170,5 @@ class FilteredTraitObserver:
                 match_func=self.filter,
                 optional=False,
             ),
-            children=[graph],
+            branches=[graph],
         )
--- a/traits/observation/_named_trait_observer.py
+++ b/traits/observation/_named_trait_observer.py
@@ -196,5 +196,5 @@ class NamedTraitObserver:
                 match_func=lambda name, trait: name == self.name,
                 optional=self.optional,
             ),
-            children=[graph],
+            branches=[graph],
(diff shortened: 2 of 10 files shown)


kitchoi:
The last commit is tagged `experimental/recursion-observe`

PR #1245, opened 2020-07-18
[WIP] Test only matched lines

rahulporuri:
Instead of testing against all of the lines in the file, parse and check only the lines that are regex matched. I've opened this PR to check on travis what kind of time difference we see when we run the entire package.


rahulporuri:
in the most recent job that succeeded, flake8 took 11.82 seconds as reported by travis - https://travis-ci.org/github/enthought/traits/jobs/707784496

PR #1276, opened 2020-08-13
FIX : Pin sphinx version to get docs to build

### LINKED ISSUE #1273: Documentation build failure on CI
Documentation build on master is failing with the following error:

```
Running Sphinx v3.2.0
making output directory... done
loading intersphinx inventory from https://docs.enthought.com/pyface/objects.inv...
loading intersphinx inventory from https://docs.enthought.com/traitsui/objects.inv...
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 43 source files that are out of date
updating environment: [new config] 43 added, 0 changed, 0 removed
reading sources... [ 13%] traits_api_reference/ctraits                         
Extension error:
Handler <function _process_docstring at 0x04720B28> for event 'autodoc-process-docstring' threw an exception (exception: <method 'trait_items_event' of 'traits.ctraits.CHasTraits' objects> is not a module, class, method, function, traceback, frame, or code object)
```

If I downgrade to Sphinx 3.1.2, then the build passes.
Currently this is causing CI builds for new pull requests to fail.

rahulporuri:
For reasons which i'm not sure of, sphinx 2.3.1 works but sphinx 3.2.0 which is the latest version available on pypi doesn't work.

fixes #1273 

**Checklist**
~- [ ] Tests~

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -302,14 +302,14 @@ setuptools.setup(
     extras_require={
         "docs": [
             "enthought-sphinx-theme",
-            "Sphinx",
+            "Sphinx==2.3.1",
         ],
         "test": [
             "Cython",
             "flake8",
             "mypy",
             "setuptools",
-            "Sphinx>=2.1.0",
+            "Sphinx==2.3.1",
             # Python 3.9 exclusions:
             #
             # * NumPy installation fails on Python 3.9 on the Ubuntu Xenial
--- a/setup.py
+++ b/setup.py
@@ -302,14 +302,14 @@ setuptools.setup(
     extras_require={
         "docs": [
             "enthought-sphinx-theme",
-            "Sphinx==2.3.1",
+            "Sphinx!=3.2.0",
         ],
         "test": [
             "Cython",
             "flake8",
             "mypy",
             "setuptools",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -302,14 +302,14 @@ def get_long_description():
...
-            "Sphinx",
+            "Sphinx==2.3.1",
         ],
         "test": [
             "Cython",
             "flake8",
             "mypy",
             "setuptools",
-            "Sphinx>=2.1.0",
+            "Sphinx==2.3.1",
```
kitchoi, line 312: I think I was the one who suggested downgrading sphinx, but I think we don't need to go all the way back to `2.3.1`, e.g. the last build with sphinx 3.1.2 does work fine.

And to stay current with less effort, do you think we could do `"Sphinx!=3.2.0"` instead? Then when there is a bug fix release, we will be able to install that. Same for the EDM dependency.

kitchoi: (EDM does not have 3.2.0 yet, but it won't mind a `!=3.2.0` as long as it can be satisfied.)

PR #1278, opened 2020-08-14
Fix containment check relying on ability to check for equality

### LINKED ISSUE #1277: Equality testing issue in traits.observation
The following code:
```
import numpy
from traits.api import HasTraits, ComparisonMode, Array, List, Union, observe
from traits.observation.api import trait

class Example(HasTraits):

    list_or_array = Union(List, Array, comparison_mode=ComparisonMode.identity)

    @observe(trait("list_or_array").list_items(optional=True))
    def something_changed(self, event):
        print(event)

example = Example()
example.list_or_array = [1, 2, 3, 4, 5]
example.list_or_array.append(5)
example.list_or_array = numpy.array([1, 2, 3, 4, 5])
```
fails on the last line with the following output
```
TraitChangeEvent(object=<__main__.Example object at 0x10190eaf0>, name='list_or_array', old=None, new=[1, 2, 3, 4, 5])
ListChangeEvent(object=[1, 2, 3, 4, 5, 5], index=5, removed=[], added=[5])
TraitChangeEvent(object=<__main__.Example object at 0x10190eaf0>, name='list_or_array', old=[1, 2, 3, 4, 5, 5], new=array([1, 2, 3, 4, 5]))
  File "/Users/cwebster/src/scratch/observe_array_issue.py", line 21, in <module>
    example.list_or_array = numpy.array([1, 2, 3, 4, 5])
  File "/Users/cwebster/.edm/envs/test-data-view-basics/lib/python3.6/site-packages/traits/observation/_observer_change_notifier.py", line 152, in __call__
    dispatcher=self.dispatcher,
  File "/Users/cwebster/.edm/envs/test-data-view-basics/lib/python3.6/site-packages/traits/observation/_has_traits_helpers.py", line 164, in observer_change_handler
    if event.new not in UNOBSERVABLE_VALUES:
```

This is because equality testing for arrays doesn't return `True` or `False`.

The fix here is to use identity testing, since all of the `UNOBSERVABLE_VALUES` are unique values like `None`.  Something like:
```
all(event.new is not value for value in UNOBSERVABLE_VALUES)
```
should work.  A similar change would need to be made on line 71 and 92 of the same file.

kitchoi:
Fixes #1277

`Undefined`, `Uninitialized`, `None` are sentinel values we should not observe. They are skipped in on_trait_change using identity check as well. This avoids errors like the one in #1277.

**Checklist**
- [x] Tests

### DIFF
--- a/traits/observation/_has_traits_helpers.py
+++ b/traits/observation/_has_traits_helpers.py
@@ -68,7 +68,7 @@ def iter_objects(object, name):
         skipped values.
     """
     value = object.__dict__.get(name, Undefined)
-    if value not in UNOBSERVABLE_VALUES:
+    if all(value is not skipped for skipped in UNOBSERVABLE_VALUES):
         yield value
 
 
@@ -89,7 +89,7 @@ def observer_change_handler(event, graph, handler, target, dispatcher):
     dispatcher : callable
         Callable for dispatching the handler.
     """
-    if event.old not in UNOBSERVABLE_VALUES:
+    if all(event.old is not skipped for skipped in UNOBSERVABLE_VALUES):
         try:
             add_or_remove_notifiers(
                 object=event.old,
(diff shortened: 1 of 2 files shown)


mdickinson (approved):
LGTM

PR #1283, opened 2020-08-27
logger.warn is deprecated, use logger.warning instead

rahulporuri:
Ref : https://docs.python.org/3.6/library/logging.html#logging.Logger.warning

**Checklist**
- [ ] ~Tests~


mdickinson (approved):
Thanks. LGTM.

PR #1293, opened 2020-09-19
Add a shell click command

rahulporuri:
This PR adds a new `shell` click command - which when run activates the development environment for traits. The usual command line options are available to choose the environment name and the runtime version of the development environment. Underneath the hood, the click command just uses the `edm shell` command with the appropriate environment name, which is why the docstring for the click command is _almost_ the same as that of `edm shell`.

**Checklist**
- [ ] ~Tests~


mdickinson (approved):
LGTM

PR #1331, opened 2020-10-27
Add script to benchmark performance of observe to compare with on_trait_change

### LINKED ISSUE #1325: Benchmark performance of observe to compare with on_trait_change
This issue is for performing benchmarks to compare the performance of on_trait_change and observe. The motivation here is to be able to identify cases when the use of observe may present a slow-down compared to on_trait_change, so that we can focus optimization efforts there in order to eliminate any barrier to the adoption of the new framework.

**Scenario 1**
Given a subclass of HasTraits with a trait defined, e.g. `name = Str()`,
compare having a method that is decorated with `@on_trait_change("name")` to without.
compare having a method that is decorated with `@observe("name")` to without.

**Scenario 2**
Given a subclass of HasTraits with a trait referring to an instance of HasTraits, e.g. `child = Instance(AnotherClass)` where `AnotherClass` has a `name` trait.
compare having a method that is decorated with `@on_trait_change("child.name")` to without.
compare having a method that is decorated with `@observe("child.name")` to without.

(3) Time for reassigning `child` to a new value.
(4) Time for reassigning `child.name` to a new value.

**Scenario 3**
Given a subclass of HasTraits with a trait that is defined as Property that depends on a simple trait
comparing having the `Property` to be defined as `Property(depends_on="name")` to without `depends_on`.
comparing having the `Property` to be defined using `Property(observe="name")` to without `observe`. (Note that this is a feature currently available on master only).

**Scenario 4**
Given a subclass of HasTraits with a trait that is defined as Property that depends on an extended trait
comparing having the `Property` to be defined as `Property(depends_on="child.name")` to without `depends_on`.
comparing having the `Property` to be defined using `Property(observe="child.name")` to without `observe`. (Note that this is a feature currently available on master only).

(3) Time for reassigning `child` to a new value.
(4) Time for reassigning `child.name` to a new value.

aaronayres35:
closes #1325 

**Checklist**
~- [ ] Tests~

### DIFF
--- a/traits/tests/check_observe_timing.py
+++ b/traits/tests/check_observe_timing.py
@@ -1,0 +1,313 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+"""
+Perform timing tests to compare the performance of on_trait_change and observe.
+
+For details on motivation and the varying "scenarios" see the related issue:
+enthought/traits#1325
+"""
+
+import timeit
+
+
+# Number of iterations to perform:
+N = 10000
+
+
+base_setup = """
+from traits.api import (
+    HasTraits, observe, on_trait_change, Str, Instance, Property
+)
+"""
+
+person_construction_template = """
+{decorator}
+class Person(HasTraits):
+    name=Str()
+"""
+
+person_instantiation_setup_template = base_setup + person_construction_template
+instantiate_person = """
+Person()
+"""
+
+person_name_reassignment_setup_template = person_instantiation_setup_template \
+                                          + "a_person = Person()"
+reassign_name = """
+a_person.name = ''
+"""
+
+
+def scenario1(decorator):
+    construct_person = person_construction_template.format(decorator=decorator)
+    construction_time = timeit.timeit(
+        construct_person, base_setup, number=N
+    )
+
+    instantiate_setup = person_instantiation_setup_template.format(
+        decorator=decorator
+    )
+    instantiation_time = timeit.timeit(
+        instantiate_person, instantiate_setup, number=N
+    )
+
+    reassign_setup = person_name_reassignment_setup_template.format(
+        decorator=decorator
+    )
+    reassign_person_name_time = timeit.timeit(
+        reassign_name, reassign_setup, number=N
+    )
+
+    return (construction_time, instantiation_time, reassign_person_name_time)
+
+
+construct_parent_setup = base_setup \
+                         + person_construction_template.format(decorator="")
+parent_construction_template = """
+{decorator}
+class Parent(HasTraits):
+    child = Instance(Person)
+"""
+
+parent_instantiation_setup_template = construct_parent_setup \
+                                      + parent_construction_template
+instantiate_parent = """
+Parent()
+"""
+
+child_reassignment_setup_template = parent_instantiation_setup_template \
+                                    + "a_parent = Parent(child=Person())"
+reassign_child = """
+a_parent.child = Person()
+"""
+
+reassign_child_name = """
+a_parent.child.name = ''
+"""
+
+
+def scenario2(decorator):
+    construct_parent = parent_construction_template.format(
+        decorator=decorator
+    )
+    construction_time = timeit.timeit(
+        construct_parent, construct_parent_setup, number=N
+    )
+
+    instantiate_parent_setup = parent_instantiation_setup_template.format(
+        decorator=decorator
+    )
+    instantiation_time = timeit.timeit(
+        instantiate_parent, instantiate_parent_setup, number=N
+    )
+
+    reassign_child_setup = child_reassignment_setup_template.format(
+        decorator=decorator
+    )
+    reassign_child_time = timeit.timeit(
+        reassign_child, reassign_child_setup, number=N
+    )
+
+    reassign_child_name_time = timeit.timeit(
+        reassign_child_name, reassign_child_setup, number=N
+    )
+
+
+    return (construction_time, instantiation_time, reassign_child_time, reassign_child_name_time)
+
+
+person_with_property_construction_template = """
+class Person(HasTraits):
+    name=Str()
+    a_property = Property({})
+"""
+
+person_with_property_instantiation_setup_template = base_setup \
+    + person_with_property_construction_template
+
+dependee_name_reassignment_setup_template = person_with_property_instantiation_setup_template + "a_person = Person()"
+
+
+def scenario3(property_args):
+    construct_person_with_property = person_with_property_construction_template.format(property_args)
+    construction_time = timeit.timeit(
+        construct_person_with_property, base_setup, number=N
+    )
+
+    instantiate_person_with_property_setup = person_with_property_instantiation_setup_template.format(property_args)
+    instantiation_time = timeit.timeit(
+        instantiate_person, instantiate_person_with_property_setup, number=N
+    )
+
+    reassign_dependee_name_setup = dependee_name_reassignment_setup_template.format(property_args)
+    reassign_dependee_name_time = timeit.timeit(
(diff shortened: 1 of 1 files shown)


aaronayres35:
Currently the script can be run and the benchmarks are all printed.  The code itself is in a bit of an ugly state though.  Theres repeated code, long/bad names, no docstrings etc.  I still need to spend some time cleaning this up

aaronayres35:
Maybe each scenarios benchmark should be placed in a separate script?  It might make each script easier to read, but it also might be nice to keep it as is to be able to run them all at once and see all the results together.  Im not sure

traits/tests/check_observe_timing.py
```diff
@@ -0,0 +1,430 @@
...
+        reassign_child_name_time
+    )
+
+
+# Code template strings to be timed for scenario 3
+
+person_with_property_construction_template = """
+class Person(HasTraits):
+    name=Str()
+    a_property = Property({})
```
kitchoi, line 175: A property without a getter is a rather unrealistic setup. 
I don't think it would change the answer, but to put that wonder to rest, let's add a getter for this.
e.g.
```
def _get_a_property(self):
    return self.name
```

traits/tests/check_observe_timing.py
```diff
@@ -0,0 +1,430 @@
...
+
+    return (construction_time, instantiation_time, reassign_dependee_name_time)
+
+
+# Code template strings to be timed for scenario 4
+
+parent_with_property_construction_template = """
+class Parent(HasTraits):
+    child = Instance(Person)
+    a_property = Property({})
```
kitchoi, line 228: Same here, let's add a getter for this.

kitchoi (commented):
Thank you!

As we discussed on a separate channel, we probably want to exercise the logic involving `@cached_property` for https://github.com/enthought/traits/issues/1316, maybe as an additional scenario. I believe the performance is quite different compared to without `@cached_property`.

kitchoi:
> Maybe each scenarios benchmark should be placed in a separate script? It might make each script easier to read, but it also might be nice to keep it as is to be able to run them all at once and see all the results together. Im not sure

That could work. An alternative to this maybe to wrap each scenario into a unittest.TestCase, it has the added benefit of this being exercised by the test suite, keeping the script up-to-date.

Oh, I forgot to mention, it might be useful to add the relative time compared to the "control" / "baseline", as the absolute value is going to be different on different machine, but the percentage difference is one useful metric people can use to communicate changes.

kitchoi:
> An alternative to this maybe to wrap each scenario into a unittest.TestCase, it has the added benefit of this being exercised by the test suite, keeping the script up-to-date.

Ah but one slight tricky thing with this approach is the last "print-to-console" step: we would want that step to be skipped when running the test suite, but not when we run the script. Feel free to ignore me on this one, this could be a future enhancement.

traits/tests/check_observe_timing.py
```diff
@@ -0,0 +1,430 @@
...
+    HasTraits, Instance, observe, on_trait_change, Property, Str
+)
+from traits.observation.api import trait
+"""
+
+person_construction_template = """
+class Person(HasTraits):
+    name = Str()
+    {decorator}
+    def some_method(self):
```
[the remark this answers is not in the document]
aaronayres35: ah yes I saw an error for this with the `Parent` class and made the change there, but missed it here. Im not sure why the same error was not showing up for this case, but in any cases I made the change now.

aaronayres35:
> > An alternative to this maybe to wrap each scenario into a unittest.TestCase, it has the added benefit of this being exercised by the test suite, keeping the script up-to-date.
> 
> Ah but one slight tricky thing with this approach is the last "print-to-console" step: we would want that step to be skipped when running the test suite, but not when we run the script. Feel free to ignore me on this one, this could be a future enhancement.

I think I will leave this as a future enhancement for now

traits/tests/check_observe_timing.py
```diff
@@ -0,0 +1,721 @@
...
+        (1) Time to import the module, i.e. the class construction.
+        (2) Time to instantiate the HasTraits object.
+        (3) Time for reassigning the trait to a new value.
+
+    Times are reported as:
+    (#): absolute_time  relative_time_to_control
+        """
+    )
+    benchmark_template = ("(1): {0:.5f} {3:5.2f}  (2): {1:.5f} {4:5.2f}  "
+                          "(3): {2:.5f} {5:5.2f}")
```
kitchoi, line 360: Nit: It can get misaligned in some situations:
```
control -
(1): 1.12156 1.00  (2): 0.03318 1.00  (3): 0.01429 1.00  (4): 0.00063 1.00
depends_on='child.name' -
(1): 1.14459 1.02  (2): 0.84582 25.49  (3): 0.18172 12.72  (4): 0.00079 1.25
```

For example, changing `{1:5.2f}` to `{1:<10.2f}` will ensure the entry fills up the space for, say 10 characters (I made up the '10' here), that would also help better separating the relative number from the absolute time visually.

kitchoi (commented):
Thank you. I think the code being timed looks good. Some comments on readability and maintainability.

mdickinson:
Closing and re-opening to trigger travis-ci.com run after migration from travis-ci.org.

traits/tests/check_observe_timing.py
```diff
@@ -0,0 +1,721 @@
...
+    a_person = Person()
+""")
+
+# Update the trait being observed
+REASSIGN_NAME = textwrap.dedent("""
+    a_person.name = ''
+""")
+
+
+def scenario1(decorator):
```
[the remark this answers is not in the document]
aaronayres35: I am going to hold off on the 'wrapping-the-scenario-into-a-class' idea for now.  I may come back to it soon, and it should be easier now after the latest round of clean up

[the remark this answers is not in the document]
kitchoi: > I am going to hold off on the 'wrapping-the-scenario-into-a-class' idea for now. I may come back to it soon, and it should be easier now after the latest round of clean up

Certainly.  I only meant to put it out there :)

traits/tests/check_observe_timing.py
```diff
@@ -0,0 +1,568 @@
...
+        formatted way
+    get_time : function
+        The function used to get the benchmark times for the current benchmark
+        scenario
+    get_time_args : list of tuples
+        The list of tuples containing the arguments to be passed the the
+        get_time function.  Note the first argument should give specifics about
+        the case being timed and will be printed as part of the report.  e.g.
+        ("@observe('name')"), or ("depends_on='name'", '@cached_property').  A
+        first argument of '' will be printed in the report as 'control'. The
```
kitchoi, line 459: Instead of saying `""` indicates control (it is a bit of a magical value), we can say the first item in the list will be used as reference / control. The use of first is still a bit magical, but it will at least decouple the concept of 'control' with the specific argument values.

traits/tests/check_observe_timing.py
```diff
@@ -0,0 +1,568 @@
...
+        first tuple should correspond to the control case.
+    """
+
+    print(description)
+
+    control_times = get_time(*get_time_args[0])
+
+    for args in get_time_args:
```
kitchoi, line 467: To enable the decoupling mentioned above, you can do:
```
for index, args in enumerate(get_time_args):
    if index == 0:
        print("control -")
        case_times = control_times
    else:
        print(args[0] + "-")
        case_times = get_time(*args)
```

traits/tests/check_observe_timing.py
```diff
@@ -0,0 +1,568 @@
...
+
+        print(benchmark_template.format(*case_times, *relative_times))
+
+    print('-' * 80)
+
+
+def main():
+    BENCHMARK_TEMPLATE_3TIMES = ("(1): {0:.5f} {3:>6.2f} "
+                                 "(2): {1:.5f} {4:>6.2f} "
+                                 "(3): {2:.5f} {5:<6.2f}")
```
kitchoi, line 487: Nit: Is this `{5:<6.2f}` instead of `{5:>6.2f}` (more consistent with others) deliberate?

kitchoi (approved):
Thank you!
LGTM. Some minor suggestions.

traits/tests/check_observe_timing.py
aaronayres35: oops, nope it was a typo

aaronayres35:
Note the benchmarking script from this PR provides relevant information about #1316

PR #1400, opened 2021-01-15
Doc: Describe Property depends_on / observe without caching context in User Manual

### LINKED ISSUE #1318: Doc: Describe Property depends_on / observe without caching context in User Manual
Currently in the user manual, the only mention of `depends_on` (or `observe`, on master) on `Property` is written in the context of a cached property. It could be misleading: one may think the `depends_on` / `observe` is not useful unless one also uses `@cached_property`, but that's not true.

(The `depends_on` / `observe` is mentioned in the traits.traits.Property API documentation without the caching context, though.)

aaronayres35:
closes #1318 

This PR pulls some content from the `Property` function docstring into the user manual in the section directly before the discussion on caching.  It still mentions the use of observe metadata in the caching section, but with slightly less bulk to try to avoid being too repetitive.  

Also, in PR #1350, the currentmodule was reset to be `traits.HasTraits` so that the Property Factory function was correctly documented.  However, the `Property` factory function actually lives in `traits.traits`, so that is updated here as well.

**Checklist**
- ~[ ] Tests~

### DIFF
--- a/docs/source/traits_user_manual/advanced.rst
+++ b/docs/source/traits_user_manual/advanced.rst
@@ -1150,7 +1150,7 @@ situations:
 You can use the **observe** metadata attribute to indicate that the
 property depends on the value of another trait. The value of **observe**
 follows the same signature as the **expression** parameter in
-|HasTraits.observe|, which is also described in
+|HasTraits.observe|, which is also described in the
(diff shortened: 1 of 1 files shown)


mdickinson (approved):
LGTM. Thank you!

PR #1421, opened 2021-01-22
Fix issues with tutorial CSS

### LINKED ISSUE #1420: Improve styling for the Traits tutorial
The CSS used for styling the appearance of the Traits tutorial in the `etsdemo` app could use some work. See discussion in #1418 and #1419.

corranwebster:
This fixes the distinct colours in the code samples, plus switches from the orange color scheme to a more usual Enthought blue.

This is a follow on from #1419 and should fix #1420.

### DIFF
--- a/traits/examples/introduction/default.css
+++ b/traits/examples/introduction/default.css
@@ -1,36 +1,36 @@
 body { background-color: #FFFFFF; }
 
-h1 { font-family: Arial; 
-     font-size: 14pt;
-     color: #303030;
-     background-color: #FCD062;
-     padding-top: 3px; 
+h1 { font-family: Arial;
+     font-size: 18pt;
+     color: #272a32;
+     background-color: #8fbfde;
+     padding-top: 3px;
      padding-left:8px;
-     padding-bottom: 3px; 
+     padding-bottom: 3px;
      padding-right: 8px; }
 
 h2 { font-family: Arial;
-     font-size: 12pt;
-     color: #303030;
-     background-color: #FCD062;
-     padding-top: 3px; 
+     font-size: 14pt;
+     color: #272a32;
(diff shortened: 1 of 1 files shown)


mdickinson:
@corranwebster Can you say what steps you took to get that screenshot?

Here's what I'm seeing:

Screenshot 2021-01-22 at 11 56 18

Steps to reproduce on my machine (from top level of the traits repository):
```
$ python etstool.py install
$ python etstool.py shell
$ pip install etsdemo
$ etsdemo
```

corranwebster:
I'll set up a clean environment and test; I did it in my default `edm` environment.

corranwebster:
I can reproduce the issue with a clean install.  Not sure what's going on, but I think it can be fixed by requesting a monospaced font.

corranwebster:
Hmmm... now I get this:

image

I think it might be worth dumping out the generated HTML - it feels like docutils is doing something strange.

corranwebster:
OK, now I am suspicious of Qt or the HTML editor doing something weird.  The output from docutils looks fairly straightforward.

corranwebster:
Running against PySide2 installed with pip gives:
image

Running with PyQt4 also looks like the above.

Clearly still something isn't quite right (indented text is a different size, for example) but this makes me think that the issue is with the way that Qt is doing something - maybe some default Qt stylesheet kicking in or something.

mdickinson:
@corranwebster What do you want to do with this PR? Should we merge as-is, or are there further changes pending?

corranwebster:
Let's merge as-is, because it's better, but perhaps keep #1420 open (or open another issue with similar content).

mdickinson (approved):
LGTM

PR #1468, opened 2021-06-09
Fix docstring to observe's dispatcher callable

kitchoi:
This PR fixes the docstring of the `dispatcher` argument on `traits.observation.api.observe` (not the decorator!).

The mention of dispatching onto a different thread is merely an example, the "i.e." is misleading.
This PR fixes that and also clarifies the default behavior.

**Checklist**
- ~Tests~


mdickinson (approved):
LGTM. Thank you!

PR #1556, opened 2021-09-29
Fix parameter name typo in stubs

mdickinson:
deligate -> delegate


rahulporuri (approved):
LGTM. I don't find any other instance of this misspelling in this repo.

PR #1671, opened 2022-07-17
Reveal and fix a masked test

rahulporuri:
The test name started with an underscore which is why unittest wasnt discovering it.

Once the leading underscore is removed, we noticed that the test assertion fails - because the `app_name` that the code infers by default is different.

We update the assertion to pass - with the strong assumption that the tests are run using `unittest` - which traits uses for test discovery and to run the tests.

**Checklist**
- [ ] ~Tests~

### DIFF
--- a/traits/etsconfig/tests/test_etsconfig.py
+++ b/traits/etsconfig/tests/test_etsconfig.py
@@ -362,14 +362,3 @@ class ETSConfigTestCase(unittest.TestCase):
         os.remove(path)
 
         self.assertEqual(data, result)
-
-
-# For running as an individual set of tests.
-if __name__ == "__main__":
-
-    # Add the non-default test of application_home...non-default because it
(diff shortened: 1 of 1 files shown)


rahulporuri:
The test failed as expected - 

```
FAIL: test_default_application_home (traits.etsconfig.tests.test_etsconfig.ETSConfigTestCase)
application home
  File "/opt/hostedtoolcache/Python/3.7.13/x64/lib/python3.7/site-packages/traits/etsconfig/tests/test_etsconfig.py", line 218, in test_default_application_home
    self.assertEqual(app_name, "tests")
AssertionError: 'unittest' != 'tests'
- unittest
+ tests
```

Fixing it now.

corranwebster:
> Probably we should rethink the whole way that `application_data` gets its default.

I have opened #1672 for this.

PR #1687, opened 2022-08-08
Ignore the traits-stubs build directory in .gitignore

mdickinson:
This PR updates `.gitignore` to ignore the traits-stubs build directory in addition to the main Traits build directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 # Build directories
 *.egg-info/
 /build/
+/traits-stubs/build/
 /dist/
 /docs/build/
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -8,6 +8,7 @@
 # Build directories
 *.egg-info/
 /build/
+/traits-stubs/build/
```
corranwebster, line 11: And dist as well?

corranwebster (approved):
Looks good, but probably want to ignore the `dist` subdirectory as well.

.gitignore
mdickinson: Good idea; done.

PR #1694, opened 2022-08-09
Update etstool.py for Python 3.8

corranwebster:
This makes Python 3.8 the default runtime to use, and updates the dependencies appropriately.  I think I don't need to make any changes to the github workflows for this - this change is mainly a developer convenience.


mdickinson (approved):
LGTM!

PR #1877, opened 2026-01-25
Update copyright end year

HalfWhitt:
I initially submitted #1875 when it was still 2025, but now its flake8 check points out the copyright end year needs updating.

- [ ] Tests


mdickinson:
> FWIW, I think we have the wrong approach here: we should be able to engineer things so that we merely get warnings when the end copyright year is out of date (perhaps along with a check that prevents making a _release_ with an outdated copyright), rather than having CI suddenly break at midnight Jan 1st every year.

Proof of concept here: https://github.com/enthought/envisage/pull/596

PR #1878, opened 2026-02-07
Fix trait_documenter to properly document traits of base classes

### LINKED ISSUE #1865: `trait_documenter` fails with simple inheritance
I have quite many headaches to get `traits` working properly with `sphinx` for a simple API reference.

My setup is the following: I have a module, say

``` python
from traits.api import Int, HasStrictTraits


class BaseClass(HasStrictTraits):
    """A base class with a docstring."""

    #: An attribute named foo.
    foo = Int(0)


class SubClass(BaseClass):
    """A subclass with a docstring."""

    #: Another attribute named bar.
    bar = Int(1)
```
and a very basic `sphinx` conf.py:
``` python
extensions = [
    "sphinx.ext.autodoc",
    "traits.util.trait_documenter"
]

autodoc_default_options = {
    'members': True,
    'member-order': 'bysource',
    'exclude-members': 'trait_added,trait_modified',
    'inherited-members': 'HasStrictTraits,HasTraits,CHasTraits',
    'show-inheritance': True, # False does not work, need to delete this line to deactivate!
}
```

Running `sphinx-build` then throws the warning

```
WARNING: No definition for the trait 'foo' could be found in class <class 'testmydocs.classes.SubClass'>.
  File "/home/pelling/micromamba/envs/testmydocs/lib/python3.14/site-packages/traits/util/trait_documenter.py", line 118, in add_directive_header
    definition = trait_definition(
        cls=self.parent,
        trait_name=self.object_name,
    )
  File "/home/pelling/micromamba/envs/testmydocs/lib/python3.14/site-packages/traits/util/trait_documenter.py", line 189, in trait_definition
    raise ValueError(
    ...<2 lines>...
    )
ValueError: No trait definition for 'foo' found in <class 'testmydocs.classes.SubClass'>
```

and the `SubClass.foo` trait is not displayed correctly:

Image

Is this the intended behaviour? What is the correct way to use `trait_documenter`? Any help here would be greatly appreciated.

itziakos:
Augment tests and some clean-up on the trait_documenter. Using AtributeDocumenter as the base class it fixes #1865

**Checklist**
- [x] Tests
- [x] Update type annotation hints in stub files

### DIFF
--- a/traits/util/trait_documenter.py
+++ b/traits/util/trait_documenter.py
@@ -198,8 +198,8 @@ def _get_definition_tokens(tokens):
 
     Parameters
     ----------
-    tokens : iteratable
-        An iteratable producing tokens.
+    tokens : iterable
+        An iterable producing tokens.
 
     Returns
     -------


traits/util/trait_documenter.py
```diff
@@ -208,8 +198,8 @@ def _get_definition_tokens(tokens):
 
     Parameters
     ----------
-    tokens : iterator
-        An iterator producing tokens.
+    tokens : iteratable
+        An iteratable producing tokens.
```
mdickinson:
```suggestion
    tokens : iterable
        An iterable producing tokens.
```

PR #1884, opened 2026-07-21
Pin copyright end year and add a strict, non-blocking style check

mdickinson:
Clean up copyright-year checking, following the approach from enthought/envisage#596 and enthought/envisage#607, to make it robust across year boundaries.

Previously the merge-blocking style check ran `flake8` with `flake8-ets`'s H102 rule against the *current* year, so the first PR of each new year would fail CI until every copyright header was bumped.

## Changes

- **Pin the copyright end year** in `setup.cfg` (`copyright-end-year = 2026`) as the single source of truth, used by both the regular style check and local `flake8` runs. The regular (merge-blocking) check now validates against this fixed year, so it no longer starts failing when the year rolls over.
- **Add a separate "Strict style check" workflow** that runs `flake8 --select H102 --copyright-end-year current`. This keeps out-of-date copyright headers conspicuous (so they still get updated early each year / before releases) without blocking merges — it's a distinct, non-required check.
- **Require `flake8-ets >= 1.1.0`**, the version that supports the `current` value for `--copyright-end-year`.
- **Switch `exclude` → `extend-exclude`** and add `.venv`, so local runs skip the virtualenv while keeping flake8's default excludes.

## Verification

Using an environment with `flake8-ets 1.1.0`:

- Regular check (pinned 2026): passes.
- Strict check (`--copyright-end-year current`, currently 2026): passes.
- Simulated year rollover (`--copyright-end-year 2027`): the strict check fails with `H102 Copyright end year (2026) should be 2027`, while the pinned regular check keeps CI green — confirming the intended behaviour.

### DIFF
--- a/.github/workflows/run-style-checks-strict.yml
+++ b/.github/workflows/run-style-checks-strict.yml
@@ -1,9 +1,9 @@
-name: Style check
+name: Strict style check
 
 on: [pull_request, workflow_dispatch]
 
 jobs:
-  style:
+  strict-style:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


mdickinson:
Changes look fine; merging.

PR #1887, opened 2026-07-21
Harden GitHub Actions workflows

mdickinson:
Security hardening for the GitHub Actions setup, following the reference pattern in enthought/envisage.

## Changes

- **Dependabot cooldown**: add a 7-day `cooldown` (`default-days: 7`) so freshly-published action releases aren't proposed immediately.
- **SHA-pinned actions**: replace mutable version tags with full 40-character commit hashes (keeping a `# vX.Y.Z` comment) across all workflows. Actions are pinned to the version currently in use — this is hardening, not an upgrade.

| Action | Was | Now |
|---|---|---|
| `actions/checkout` | `@v6` | `@d23441a…af803 # v6.1.0` |
| `actions/setup-python` | `@v6` | `@ece7cb06…8c61a1 # v6.3.0` |
| `pypa/cibuildwheel` | `@v3.3.1` | `@298ed2fb…dd3a7e # v3.3.1` |

Tags were resolved against the GitHub API; the `setup-python` pin matches the one used in envisage. Dependabot will continue to update these while keeping the SHA authoritative.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,5 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 7
--- a/.github/workflows/release-to-pypi.yml
+++ b/.github/workflows/release-to-pypi.yml
@@ -16,10 +16,10 @@ jobs:
     steps:
 
     - name: Check out the release commit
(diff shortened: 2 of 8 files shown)


mdickinson:
Simple change that doesn't touch business logic. Merging.

PR #1891, opened 2026-07-28
IT-6150: switch PyPI publishing to trusted publishing (OIDC)

mdickinson:
**Extension of IT-6150** — move PyPI release uploads to GitHub Actions OIDC **trusted publishing**, removing the long-lived `PYPI_USERNAME` / `PYPI_PASSWORD` secrets. IT-6150 itself covered only apptools, envisage and flake8-ets; this applies the same change to traits, which needs a restructure rather than a like-for-like swap.

## Why this one is not a drop-in change

`pypa/gh-action-pypi-publish` is Docker-based and fails fast on non-Linux runners:

```yaml
- name: Fail-fast in unsupported environments
  if: runner.os != 'Linux'
  run: |
    >&2 echo This action is only able to run under GNU/Linux environments
    exit 1
```

`build-wheels` runs on `windows-latest`, `macos-15-intel`, `macos-latest` and `ubuntu-latest`, and the uploads happened inside that matrix, so the action cannot simply replace `twine upload` in place.

## What changed (`.github/workflows/release-to-pypi.yml`)

- `build-wheels` and `build-sdist` no longer upload. They store their distributions with `actions/upload-artifact` (pinned to v7.0.1, the version already used in envisage).
- New Linux-only `publish` job with `environment: pypi` and `permissions: id-token: write`. It retrieves every `dist-*` artifact into `dist/` with `merge-multiple: true` and publishes once via `pypa/gh-action-pypi-publish` (pinned to `v1.14.1`, matching the other three repos).
- `twine check --strict` stays in the build jobs, so a bad distribution is caught before anything is uploaded.
- Two step names changed from "build and upload" to "build and check", since those steps no longer upload.

## Incidental improvements

- **One approval per release instead of five.** `environment: pypi` on the matrix job would have created a separate deployment, and so a separate review request, for each of the four wheel legs plus the sdist job.
- **No more partially-uploaded releases.** Uploads previously happened independently in five jobs, so a late failure could leave an incomplete release on PyPI with those filenames permanently burned. There is now a single upload of the complete set.
- **The dry run is preserved.** `if: github.event_name == 'release'` stays on the publish job, so `workflow_dispatch` still builds and checks without uploading. Because the job is *skipped* rather than gated, a dry run does not enter the `pypi` environment and does not request a deployment review.

## Note on `permissions`

The publish job sets only `id-token: write`, deliberately unlike the other three repos which also set `contents: read`. There is no `actions/checkout` in this job, and `actions/download-artifact` needs no elevated scope for same-run artifacts, so nothing else is required.

## GitHub-side configuration (already done)

The `pypi` environment on this repository is configured identically to apptools, envisage and ets-copyright-checker:

- Required reviewers: `en_ess-ets-internal-release` team, self-review permitted
- Deployment branch policy: tags matching `*.*.*` only

## ⚠️ Not ready to merge — validation still outstanding

The trusted publisher is configured on the PyPI side (`enthought/traits`, workflow `release-to-pypi.yml`, environment `pypi`), but the OIDC path has not yet been exercised in this repository.

Validation cannot use a plain `workflow_dispatch`, because dispatch skips the publish job by design. Proposed approach, avoiding a real GitHub release (which would notify watchers):

1. On a scratch commit, temporarily relax the publish job's `if` to also accept `workflow_dispatch`.
2. Tag it with a throwaway `*.*.*` tag and dispatch against that tag.
3. Expect the deployment review to fire, and the upload to then fail with `400 File already exists` — `setup.py` on main declares 7.1.0 with `IS_RELEASED = True`, and traits 7.1.0 is on PyPI, so no real upload can occur.
4. Delete the tag, discard the scratch commit, and merge this PR unchanged.

Once merged, `PYPI_USERNAME` and `PYPI_PASSWORD` can be deleted from this repository.


mdickinson:
Test run (failed at duplicate filenames on PyPI, as expected): https://github.com/enthought/traits/actions/runs/30359265547

PR #1895, opened 2026-08-07
Require Python >= 3.10, and replace pkg_resources with importlib.resources

mdickinson:
Raises the minimum supported Python to 3.10, and cleans up the `pkg_resources` usage found along the way.

## Version floor

Python 3.9 is well beyond end of life, and a 3.10 floor matches other recent ETS releases (Envisage, Mayavi).

Note on the resulting wheel set: cibuildwheel 4.x dropped cp38, but it still builds cp39 quite happily. What removes cp39 is the `python_requires` bump in this PR — cibuildwheel reads `python_requires` out of `setup.py` and filters its build identifiers accordingly. Verified with `cibuildwheel 4.2.0 --print-build-identifiers` against this branch:

| `requires-python` | identifiers selected |
| --- | --- |
| `>=3.8` (main) | **cp39**, cp310, cp311, cp312, cp313, cp314(t), cp315(t) |
| `>=3.10` (this PR) | cp310, cp311, cp312, cp313, cp314(t), cp315(t) |

So dropping cp39 wheels is a deliberate choice here, not something the toolchain forced.

- `python_requires` is now `>=3.10`
- `'3.8'` and `'3.9'` removed from all four workflow matrices
- `README.rst`, `intro.rst` and `.readthedocs.yaml` updated
- Dropped the "Installation from source on macOS" section, which was specifically about the system Python 3.8 on Catalina (#1337)
- `etstool.py` moves from the 3.8 EDM runtime to 3.11 (EDM offers 3.8 and 3.11, and 3.8 is unmaintained)

Cleanups the new floor allows:

- `TraitDict.__ior__` and its tests are no longer guarded on `sys.version_info >= (3, 9)`, and the `else` branch asserting `TypeError` on older Pythons is gone
- black's `target-version` was still `py36`

## pkg_resources → importlib.resources

Worth reviewing on its own merits, because this turned out to be more than cosmetic. setuptools 81 deprecated `pkg_resources` and setuptools 83 does not ship it at all. CI installs setuptools 83, so everything behind `@requires_pkg_resources` has been **silently skipping**. From the 3.15 job on #1894:

```
test_unpickling_historical_pickles ... skipped 'pkg_resources not available'
test_dunder_version ... skipped 'pkg_resources not available'
test_version_version ... skipped 'pkg_resources not available'
```

`traits.stubs_tests.test_all` and `test_etsdemo_info` were skipping for the same reason. All of them now run, and all pass.

One of these was a genuine user-facing bug rather than just a dormant test: **`traits.examples._etsdemo_info.introduction` is shipped code**, reached through the `etsdemo_data` entry point, and it did `import pkg_resources` at call time. That raises `ModuleNotFoundError` for anyone with a recent setuptools, and the only test covering it was skipped by the same missing module that broke it.

`test_version` needs a PEP 440 parser and there's no stdlib equivalent, so `packaging.version.parse` replaces `pkg_resources.parse_version` and `packaging` joins the `test` extra. Everything else moves to `importlib.resources`, using `as_file` in the stubs tests where a real filesystem path is needed for mypy.

## Verification

- Full suite on 3.10, non-editable install with `[test]`: **1628 tests, OK, 0 skips**
- Core suite on 3.15.0b2: 1628 tests, OK (91 skipped for missing optional deps — one fewer than before, since the historical-pickles test no longer needs `pkg_resources`)
- `flake8` clean
- `python -m sphinx -b html -W source build` succeeds, so the removed docs section breaks nothing

## Note on the changelog

No `CHANGES.rst` entry, following the convention that the changelog is written at release-prep time rather than per-PR. The `_etsdemo_info` fix deserves its own line under Fixes when 7.2.0 is prepared.


mdickinson:
Review notes. I checked the mechanical claims against the code and the CI logs rather than reading the diff alone.

## What checks out

- **The `__ior__` dedent is clean.** Comparing the ASTs of `TraitDict.__ior__` on `main` and on this branch, the only difference is the docstring's indentation — no accidental change to the logic. `import sys` is correctly dropped, with no remaining references in the module. Both relocated tests (`test_ior`, `test_ior_is_quiet_if_no_change`) are AST-identical to their originals.
- **The un-skipping claim is real.** The `tests (ubuntu-latest, 3.10)` job log shows `Ran 1628 tests` with zero `skipped` lines. The corresponding core job's 91 skips are now only traitsui (39), numpy (39), Sphinx (9), packaging (2) and mypy (2) — no `pkg_resources` skips remain anywhere.
- **The Read the Docs changes are correct on all three points.** RTD's v2 config reference lists `3.14` as a valid `build.tools.python` value, documents `sphinx.configuration` as `Required: true`, and no longer carries `python.version` in the schema.
- **No stale version references survive.** The only remaining 3.8/3.9 mentions are historical `CHANGES.rst` entries, correctly left alone. The classifiers list no minor versions, so there's nothing to update there.
- **`as_file` on a directory is safe at the new 3.10 floor** for filesystem installs, via the degenerate `pathlib.Path` registration; and `run_mypy` copies the file to a tempdir anyway, so no path escapes the context manager.
- **The changelog convention holds up** — `git log -- CHANGES.rst` confirms it's only touched at release-prep time.

## Findings

### 1. `packaging` is missing from two other dependency lists

This is the one gap in the PR's own thesis: the same silent-skip failure mode, in the places `setup.py` doesn't cover.

- `etstool.py`, `common_dependencies` — not listed, so `test_version` will skip in EDM runs.
- `.github/workflows/test-from-pypi.yml` — both jobs hand-roll their install list (`defusedxml numpy setuptools Sphinx traits traitsui`) and don't include it.

In both cases the two version tests will report as skipped rather than run.

### 2. `setuptools` is now dead weight in those same lists

Nothing under `traits/` imports `setuptools` or `pkg_resources` any more. Its presence in the `test` extra and in `test-from-pypi.yml` was there to make `pkg_resources` importable; the build-time requirement is separately covered by `pyproject.toml`'s `build-system.requires`.

### 3. The docs CI job and RTD now build on different Pythons

`test-documentation-build.yml` pins 3.13, while `.readthedocs.yaml` now says 3.14. The rationale in the commit message ("Sphinx 9 requires Python >= 3.12") is satisfied by 3.13 equally, and no PR check exercises the RTD config, so 3.14 is the unverified one of the pair. Worth aligning the two in whichever direction.

### 4. `test_numpy_examples` documents a parameter it doesn't take

In `traits/stubs_tests/test_all.py`, the docstring carries a `filename_suffix` Parameters section copied from `test_all`, but the signature is just `(self)`. Pre-existing, but this PR is already editing that method.

### 5. `requires_pkg_resources` was public API

`traits.testing.optional_dependencies` is published through `automodule ... :members: :undoc-members:` in `docs/source/traits_api_reference/traits.testing.rst`, so `pkg_resources` and `requires_pkg_resources` were importable public names that other ETS packages may be using. Worth a line under Removals when 7.2.0 is prepared, alongside the `_etsdemo_info` fix already noted in the PR description.

### Minor

- `str(files("traits") / "examples" / "introduction")` yields a non-existent path under zip import, where `resource_filename` would have extracted to a cache. Fine in practice given `zip_safe=False` and the C extension, but it is a behaviour change. `files("traits.examples") / "introduction"` would also be a little more direct.
- The `import importlib.resources` in `_etsdemo_info.introduction` no longer needs to be function-local now that it's stdlib rather than an optional third-party import.
- The comment in `test_version.py` now reads "Round-trip through parse", which is slightly orphaned from its subject.

🤖 Review performed with [Claude Code](https://claude.com/claude-code)

mdickinson:
Re-reviewed after 64ac92c. All five findings are addressed, and I checked the fixes rather than taking the commit message at its word.

## Verified

**1. `packaging` added to the other two dependency lists.** Since no workflow runs `etstool.py`, I checked the EDM egg name directly rather than assuming it: `edm search packaging` returns `packaging 22.0-2` and `25.0-1` from `enthought/free` for the 3.11 runtime, so `common_dependencies` will resolve.

**2. `setuptools` dropped from the `test` extra.** The 3.10 full-test job still reports `Ran 1628 tests` / `OK` with zero skip lines, so removing it reintroduced nothing. Build isolation covers the sdist case in `test-from-pypi.yml` through `build-system.requires`, so its removal there is safe too.

**3. Docs CI aligned to 3.14.** The job log confirms it genuinely ran on 3.14 and installed Sphinx 9.1.0 with enthought-sphinx-theme 0.7.5, passing under `-W`. CI and Read the Docs now build on the same Python.

**4. `test_numpy_examples` docstring.** The phantom Parameters section is gone.

**5. `files("traits.examples")` and the module-level import.** `traits/examples/__init__.py` is empty apart from the licence header, so the anchor change adds no import cost, and the module-level `importlib.resources` is safe at entry-point load time.

Full matrix is green: 36 jobs across three OSes and 3.10–3.15, plus docs, style and strict-style.

## Residual notes

The `requires_pkg_resources` removal from the documented `traits.testing.optional_dependencies` module is still outstanding, which matches the suggestion above — it wants a line under Removals at 7.2.0 release prep rather than a code change now.

Worth knowing for anyone merging: `etstool.py` and `test-from-pypi.yml` are the two changed files with no PR-level CI coverage. The EDM search above covers the first; the second only runs on a schedule, so its next scheduled run (the 8th or 22nd) is the first real exercise of that change.

Nothing new introduced by the fix commit. Looks good to me.

🤖 Review performed with [Claude Code](https://claude.com/claude-code)

mdickinson:
Transparency: all of the above was Claude Code (Opus 5) talking to itself. This is me (human, Mark) stating that I've reviewed the PR contents and it LGTM.

